Issue #42, opened 2026-03-25
fnmatch case sensitivity is platform-dependent in policy engine


benthomasson:
## Feature Request

The policy engine uses `fnmatch.fnmatch` for all pattern matching. fnmatch is case-sensitive on Linux but case-insensitive on macOS. A policy rule that works in development on macOS may silently fail to match in production on Linux (or vice versa).

## Belief

`policy-fnmatch-platform-dependent`

## Proposed Resolution

Use `fnmatch.fnmatchcase` for consistent case-sensitive matching across platforms, or document the platform dependency explicitly.

---
*Filed from ftl2-expert spec design gaps*

PR #82, opened 2026-04-11
Document shell/command/raw equivalence in policy engine

### LINKED ISSUE #72: Policy: command and raw modules bypass shell deny rules
## Summary

A policy rule denying `shell` in production does not block `command` or `raw` modules, which can execute the same destructive operations. This is a known gap documented in the PRD risks section.

## Problem

```yaml
rules:
  - decision: deny
    match:
      module: shell
      environment: prod
    reason: "Use proper modules instead of shell in production"
```

This rule blocks `shell` but the AI author can achieve the same effect with:
- `await ftl.host.command(cmd="rm -rf /")` — `command` is not `shell`
- `await ftl.host["ansible.builtin.raw"](cmd="rm -rf /")` — `raw` is not `shell`

The human stakeholder who writes this policy rule believes destructive shell access is blocked in production. It is not.

## Proposal

Options (not mutually exclusive):

1. **Documentation**: Make the bypass explicit in policy engine docs so policy authors know to deny all three modules together. Provide an example:
   ```yaml
   rules:
     - decision: deny
       match: {module: shell, environment: prod}
       reason: "No shell in prod"
     - decision: deny
       match: {module: command, environment: prod}
       reason: "No command in prod"
     - decision: deny
       match: {module: "*.raw", environment: prod}
       reason: "No raw in prod"
   ```

2. **Module groups**: Support a `module_group` match key that maps to sets of related modules (e.g., `shell_access: [shell, command, raw]`)

3. **Semantic categories**: Tag modules with categories (e.g., `arbitrary_execution`) and allow deny rules to match on category

## Context

- The policy engine uses fnmatch patterns — `shell` only matches `shell`, not `command`
- This is a semantic gap, not a technical bug — the policy engine works as designed
- For AI safety, the gap matters because the AI author may not distinguish between `shell` and `command` when generating automation

benthomasson:
## Summary

- Add explicit warning to README policy example that denying `shell` alone does not block `command` or `raw`
- Fix example policies to deny all three equivalent modules consistently
- Add docstring warning to `Policy` class about module equivalence

## Changes

| File | Change |
|------|--------|
| README.md | Expanded policy example to deny all three modules; added note |
| policy.py | Added warning to Policy class docstring |
| development-permissive.yaml | Added `command` deny for `rm -rf /*` |
| layered/00-base.yaml | Added `command` deny for `rm -rf /*` |
| layered/10-production.yaml | Added `*.raw` deny (was missing vs flat example) |

## Test plan

- [x] All 86 policy tests pass
- [ ] Verify README renders correctly on GitHub

Closes #72

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,25 +86,11 @@ Restrict what actions are permitted based on module, host, environment, and para
 ```yaml
 # policy.yml
 rules:
-  # IMPORTANT: shell, command, and raw can all execute arbitrary commands.
-  # Denying only one leaves the others as bypass routes.
   - decision: deny
     match:
       module: "shell"
       environment: "prod"
     reason: "Use proper modules in production"
-
-  - decision: deny
-    match:
-      module: "command"
-      environment: "prod"
-    reason: "Use proper modules in production"
-
-  - decision: deny
-    match:
-      module: "*.raw"
-      environment: "prod"
-    reason: "Raw module execution not permitted in production"
 
   - decision: deny
     match:
@@ -114,9 +100,9 @@ rules:
     reason: "No destructive actions on production hosts"
 ```
 
-> **Note:** The `shell`, `command`, and `raw` modules can all execute arbitrary commands.
-> A policy that denies only `shell` does not block `command` or `raw` — always deny all three
-> together when restricting arbitrary execution. See `examples/policies/` for reference.
+> **Note:** The `shell`, `command`, and `raw` modules are treated as equivalent by the
+> policy engine.  A rule denying `shell` automatically blocks `command` and `raw` too,
+> including FQCN variants like `ansible.builtin.raw`.
 
 ```python
 async with automation(policy="policy.yml", environment="prod") as ftl:
--- a/README.md
+++ b/README.md
@@ -86,6 +86,8 @@ Restrict what actions are permitted based on module, host, environment, and para
 ```yaml
 # policy.yml
 rules:
+  # IMPORTANT: shell, command, and raw can all execute arbitrary commands.
+  # Denying only one leaves the others as bypass routes.
   - decision: deny
     match:
       module: "shell"
@@ -94,11 +96,27 @@ rules:
 
   - decision: deny
     match:
+      module: "command"
+      environment: "prod"
+    reason: "Use proper modules in production"
+
+  - decision: deny
+    match:
+      module: "*.raw"
+      environment: "prod"
+    reason: "Raw module execution not permitted in production"
+
+  - decision: deny
+    match:
       module: "*"
       param.state: "absent"
       host: "prod-*"
     reason: "No destructive actions on production hosts"
 ```
+
+> **Note:** The `shell`, `command`, and `raw` modules can all execute arbitrary commands.
+> A policy that denies only `shell` does not block `command` or `raw` — always deny all three
+> together when restricting arbitrary execution. See `examples/policies/` for reference.
 
 ```python
 async with automation(policy="policy.yml", environment="prod") as ftl:
--- a/examples/policies/development-permissive.yaml
+++ b/examples/policies/development-permissive.yaml
@@ -1,17 +1,10 @@
 # Development — minimal restrictions, mostly a safety net.
 #
-# IMPORTANT: shell, command, and raw can all execute arbitrary commands.
-# Denying only "shell" leaves "command" and "raw" as bypass routes.
-# Always deny all three together when restricting arbitrary execution.
+# The policy engine treats shell, command, and raw as equivalent.
+# Denying any one automatically blocks the others.
 rules:
   - decision: deny
     match:
       module: shell
-      param.cmd: "rm -rf /*"
-    reason: "Recursive root deletion is never permitted"
-
-  - decision: deny
-    match:
-      module: command
       param.cmd: "rm -rf /*"
     reason: "Recursive root deletion is never permitted"
--- a/examples/policies/development-permissive.yaml
+++ b/examples/policies/development-permissive.yaml
@@ -1,7 +1,17 @@
 # Development — minimal restrictions, mostly a safety net.
+#
+# IMPORTANT: shell, command, and raw can all execute arbitrary commands.
+# Denying only "shell" leaves "command" and "raw" as bypass routes.
+# Always deny all three together when restricting arbitrary execution.
 rules:
   - decision: deny
     match:
       module: shell
+      param.cmd: "rm -rf /*"
+    reason: "Recursive root deletion is never permitted"
+
+  - decision: deny
+    match:
+      module: command
       param.cmd: "rm -rf /*"
     reason: "Recursive root deletion is never permitted"
--- a/examples/policies/layered/00-base.yaml
+++ b/examples/policies/layered/00-base.yaml
@@ -1,19 +1,12 @@
 # Base safety rules — universal, loaded first.
 # Name with 00- prefix so it sorts before environment-specific files.
 #
-# IMPORTANT: shell, command, and raw can all execute arbitrary commands.
-# Denying only "shell" leaves "command" and "raw" as bypass routes.
-# Always deny all three together when restricting arbitrary execution.
+# The policy engine treats shell, command, and raw as equivalent.
+# Denying any one automatically blocks the others.
 rules:
   - decision: deny
     match:
       module: shell
-      param.cmd: "rm -rf /*"
-    reason: "Recursive root deletion is never permitted"
-
-  - decision: deny
-    match:
-      module: command
       param.cmd: "rm -rf /*"
     reason: "Recursive root deletion is never permitted"
 
--- a/examples/policies/layered/00-base.yaml
+++ b/examples/policies/layered/00-base.yaml
@@ -1,9 +1,19 @@
 # Base safety rules — universal, loaded first.
 # Name with 00- prefix so it sorts before environment-specific files.
+#
+# IMPORTANT: shell, command, and raw can all execute arbitrary commands.
+# Denying only "shell" leaves "command" and "raw" as bypass routes.
+# Always deny all three together when restricting arbitrary execution.
 rules:
   - decision: deny
     match:
       module: shell
+      param.cmd: "rm -rf /*"
+    reason: "Recursive root deletion is never permitted"
+
+  - decision: deny
+    match:
+      module: command
       param.cmd: "rm -rf /*"
     reason: "Recursive root deletion is never permitted"
 
--- a/examples/policies/layered/10-production.yaml
+++ b/examples/policies/layered/10-production.yaml
@@ -16,6 +16,12 @@ rules:
 
   - decision: deny
     match:
+      module: "*.raw"
+      environment: prod
+    reason: "Raw module execution not permitted in production"
+
+  - decision: deny
+    match:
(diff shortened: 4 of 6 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#82
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### src/ftl2/policy.py:_MODULE_EQUIVALENCE_GROUPS and _MODULE_EQUIVALENTS
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The pre-computed lookup is clean — built at module import time, maps each short name to its full equivalence group. The `frozenset` usage is appropriate for O(1) membership checks. Minor nit: loop variables `_group` and `_member` leak into module namespace, but the underscore prefix convention is sufficient.

### src/ftl2/policy.py:_module_matches
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The logic is correct for the cases tested, but there's a gap: **FQCN patterns in the rule are not equivalence-expanded**. If a policy author writes `module: "ansible.builtin.shell"`, the lookup `_MODULE_EQUIVALENTS.get("ansible.builtin.shell")` returns `None` because only short names are keys. This means `command` and `raw` would NOT be blocked, silently breaking the equivalence guarantee. The function should extract the short name from the *pattern* too (not just the module_name) before checking equivalence. No test covers this case. Also untested: a FQCN rule pattern like `ansible.builtin.shell` evaluated against FQCN module `ansible.builtin.command` — this would fail to match despite equivalence.

### src/ftl2/policy.py:Policy._matches
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean delegation to `_module_matches` for the `module` key. The change is minimal and surgical — only the module matching line was changed, preserving all other match logic.

### tests/test_policy.py:TestModuleEquivalenceGroups
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Good coverage of the core scenarios: all pairwise equivalences, FQCN module names, environment interaction, unrelated modules, glob bypass. Missing tests: (1) FQCN *pattern* in rule (e.g. `module: "ansible.builtin.shell"`) — does equivalence fire? (it doesn't, and that's likely a bug); (2) case sensitivity interaction with equivalence (e.g. rule `module: "Shell"` should NOT trigger equivalence for `command`); (3) `param.*` conditions combined with equivalence (e.g. deny `shell` with `param.cmd: "rm *"` — does `command` with that param also get denied?).

### Self-Review
**Limitations:** Did not verify whether any caller in the codebase passes FQCN as the module pattern in policy rules (would confirm whether the FQCN-pattern gap is theoretical or practical). Did not check whether the existing `TestPolicyCaseSensitivity` tests still pass with the equivalence changes (e.g., `test_module_pattern_is_case_sensitive` tests `Shell` vs `shell` — equivalence shouldn't interfere, but didn't run the tests to confirm).

### Feature Requests
- Include related existing tests that test the same source functions so I can verify no regressions from signature/behavior changes
- Show `git log --oneline` for the commits being reviewed to understand ordering and intent

## gemini [BLOCK]

### README.md
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** MISSING

The documentation added explicitly claims that "A policy that denies only `shell` does not block `command` or `raw` — always deny all three together". This directly contradicts the new code in `policy.py` which implements module equivalence specifically so that a rule denying `shell` DOES automatically block `command` and `raw`. The new documentation is factually incorrect with respect to the new code changes.

### examples/policies/development-permissive.yaml
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** MISSING

The added comments state that denying only "shell" leaves "command" and "raw" as bypass routes. Due to the new equivalence logic in `policy.py`, this is no longer true. The explicitly added rules for `command` and `raw` are technically redundant but functionally harmless, though the comments will mislead users.

### examples/policies/layered/00-base.yaml
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** MISSING

Similar to the development policy, the added comments and explicit deny rules for `command` contradict the new equivalence feature implemented in `policy.py`.

### examples/policies/layered/10-production.yaml
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** MISSING

The added rules for `*.raw` are redundant due to the new module equivalence logic.

### src/ftl2/policy.py
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `_module_matches` implementation introduces a severe inconsistency in how FQCNs are handled. Because the equivalence check extracts the short name from the target `module_name`, a policy rule matching exactly `"shell"` will successfully block `"ansible.builtin.shell"`. However, for modules not in an equivalence group (like `"ping"`), a policy rule matching exactly `"ping"` will NOT block `"ansible.builtin.ping"` (since it falls back to strict `fnmatch`). This means FQCN matching behavior changes drastically depending on whether a module happens to be in an equivalence group. Additionally, the implementation breaks if a user tries to use globbing like `*.shell` to handle FQCNs consistently, as globs completely bypass the equivalence logic and will fail to block `command` or `raw`.

### tests/test_policy.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The tests accurately cover the implemented logic, correctly validating the equivalence behavior and the glob bypass. However, they fail to highlight the major inconsistency in FQCN matching between equivalence group modules and normal modules.

### Self-Review
**Limitations:** None. The diff and context provided everything necessary to identify the logical flaws, inconsistent behaviors, and documentation contradictions.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#82
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/policy.py:_MODULE_EQUIVALENCE_GROUPS and _MODULE_EQUIVALENTS
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Module-level constants are correctly structured. The pre-computed lookup dict maps each member of a group to its full group frozenset. The loop variables `_group` and `_member` leak into module scope — cosmetically imperfect but functionally harmless in Python.

### src/ftl2/policy.py:_module_matches
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The logic is sound for the tested cases, but there are two edge cases worth noting:

1. **Case sensitivity interaction with equivalence**: The equivalence lookup uses `_MODULE_EQUIVALENTS.get(pattern_short)`, where the keys are lowercase (`shell`, `command`, `raw`). The existing `TestPolicyCaseSensitivity.test_module_pattern_is_case_sensitive` test has a rule with `match={"module": "Shell"}`. With this change, `pattern_short` would be `"Shell"` and `_MODULE_EQUIVALENTS.get("Shell")` returns `None`, so equivalence is silently skipped for mixed-case names. This is probably the desired behavior (case-sensitive matching is an explicit design goal per issue #42), but there's no test that verifies the intersection: e.g., a rule denying `"Shell"` does NOT block `"command"`. Worth adding a test to lock down this contract.

2. **FQCN with non-`ansible.builtin` namespace**: If a hypothetical collection had `custom.collection.shell`, `pattern_short` would be `"shell"` and it would trigger equivalence with `command`/`raw`. This is arguably a feature not a bug (the short name is what matters), but it could surprise users. The docstring partially addresses this ("last dotted component"), which is acceptable.

### src/ftl2/policy.py:Policy._matches
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The change is minimal — swapping `fnmatch.fnmatchcase` for `_module_matches` in the `key == "module"` branch. All other match keys are unaffected. The docstring update is accurate.

### tests/test_policy.py:TestModuleEquivalenceGroups
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Thorough test coverage: all pairwise equivalences (shell↔command, shell↔raw, command↔raw), FQCN patterns blocking short names and vice versa, glob bypass, environment interaction, and unrelated module non-interference. 14 test methods cover the feature well. Minor gap: no test for case-sensitivity interaction with equivalence (see concern above), and no test for `deny raw` blocking `ansible.builtin.shell` (the reverse FQCN direction is only tested for `shell` pattern).

### examples/policies/layered/10-production.yaml
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Removing the now-redundant `command` deny rule is correct — the `shell` deny rule now covers `command` and `raw` via equivalence. The comment explains the behavior clearly. This is a good demonstration of the feature's value.

### Self-Review
**Limitations:** - Could not verify whether `AutomationContext.execute()` or the `ModuleProxy` normalizes module names before calling `policy.evaluate()` — if module names are passed as FQCN from the proxy but the policy has short names, the equivalence code handles this correctly, but I couldn't verify the full call chain.
- No integration tests were modified or reviewed — only unit tests. The `10-production.yaml` change removes a rule, but there's no test that loads this example file and verifies end-to-end behavior.
- Did not check if there are other callers of `_matches` beyond `evaluate()` that might be affected.

### Feature Requests
- Include callers of modified functions (e.g., who calls `_matches` or `evaluate`) to verify integration impact
- Show the full test file when tests are modified, not just the diff, to spot broken existing tests more easily
- Flag when example/config files are changed but have no corresponding integration test

## gemini [PASS]

### src/ftl2/policy.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The implementation of module equivalence is clean and efficient. Computing `_MODULE_EQUIVALENTS` at module import time prevents overhead during rule evaluation. The `_module_matches` function correctly handles short names, FQCNs (Fully Qualified Collection Names), and appropriately bypasses equivalence expansion when glob patterns are used.

### tests/test_policy.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The added `TestModuleEquivalenceGroups` class is exceptionally thorough. It effectively verifies the equivalence mapping forwards and backwards across different module combinations (shell, command, raw), properly tests FQCN edge cases across namespaces, checks that other match conditions (like environment) are still respected, and confirms that glob patterns bypass the equivalence grouping.

### Self-Review
**Limitations:** No specifications or related issues were provided, so the review relies purely on the logic and self-consistency of the diff itself.

PR #83, opened 2026-04-11
Add maximum protocol message size to prevent unbounded allocation

### LINKED ISSUE #74: Add maximum protocol message size to prevent unbounded memory allocation
## Summary

`GateProtocol.read_message()` reads the 8-byte hex length prefix and then allocates a buffer of that size with no upper bound check. A malformed or malicious message with a large length prefix could cause unbounded memory allocation.

## Problem

In `message.py`, `read_message()`:

1. Reads 8-byte hex length prefix
2. Parses it as an integer (`int(length_bytes, 16)`)
3. Reads that many bytes from the stream

An 8-byte hex prefix can encode up to `0xFFFFFFFF` = 4,294,967,295 bytes (~4 GB). There is no check that the length is within a reasonable range before allocating the read buffer.

This affects both sides:
- Controller reading gate responses
- Gate reading controller commands

## Proposal

Add a maximum message size constant and check before reading:

```python
MAX_MESSAGE_SIZE = 64 * 1024 * 1024  # 64 MB — generous upper bound

length = int(length_bytes, 16)
if length > MAX_MESSAGE_SIZE:
    raise ProtocolError(f"Message size {length} exceeds maximum {MAX_MESSAGE_SIZE}")
```

The limit should be large enough for any legitimate message (module bundles can be large) but small enough to prevent memory exhaustion. The exact value should be informed by the largest legitimate message observed in practice.

## Context

- Rated MEDIUM severity in the original PRD risk assessment
- For permanent gates, the gate process is a longer-lived target
- The protocol is only exposed over SSH (not a public network), which limits the attack surface but does not eliminate it

benthomasson:
## Summary

- Add `MAX_MESSAGE_SIZE` (64 MB) constant to `message.py`
- Check message length against limit in `read_message()` after parsing the hex prefix, before allocating the read buffer
- Previously, the 8-byte hex prefix could encode up to 4 GB (`0xFFFFFFFF`), allowing unbounded memory allocation from malformed or malicious messages

## Changes

| File | Change |
|------|--------|
| `src/ftl2/message.py` | Add `MAX_MESSAGE_SIZE` constant and size check in `read_message()` |
| `tests/test_message.py` | 4 new tests: oversized rejected, at-limit allowed, max hex rejected, constant verified |

## Test plan

- [x] All 17 message protocol tests pass
- [x] Oversized length prefix (MAX+1) raises ProtocolError before reading body
- [x] At-limit message passes size check (fails on incomplete body as expected)
- [x] Maximum hex value (0xFFFFFFFF / 4 GB) rejected

Closes #74

### DIFF
--- a/src/ftl2/message.py
+++ b/src/ftl2/message.py
@@ -26,6 +26,12 @@ class ProtocolError(Exception):
     pass
 
 
+# Maximum allowed message size (64 MB).  Large enough for any legitimate
+# payload (base64-encoded module bundles, SystemMetrics, etc.) but small
+# enough to prevent memory exhaustion from a malformed length prefix.
+MAX_MESSAGE_SIZE = 64 * 1024 * 1024
+
+
 class GateProtocol:
     """Protocol handler for gate communication.
 
@@ -245,6 +251,12 @@ class GateProtocol:
                 length = int(length_hex, 16)
             except (ValueError, UnicodeDecodeError) as e:
                 raise ProtocolError(f"Invalid hex length: {length_bytes!r}") from e
+
+            if length > MAX_MESSAGE_SIZE:
+                raise ProtocolError(
+                    f"Message size {length} bytes exceeds maximum "
+                    f"({MAX_MESSAGE_SIZE} bytes)"
+                )
 
             # Read message body, skipping leading whitespace
             # (newline between length and body in interactive mode)
--- a/tests/test_message.py
+++ b/tests/test_message.py
@@ -4,7 +4,7 @@ import asyncio
 
 import pytest
 
-from ftl2.message import GateProtocol, ProtocolError
+from ftl2.message import GateProtocol, MAX_MESSAGE_SIZE, ProtocolError
 
 
 class TestGateProtocol:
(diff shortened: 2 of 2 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#83
**Models:** claude, gemini
**Gate:** [PASS] PASS

## claude [PASS]

### src/ftl2/message.py:MAX_MESSAGE_SIZE
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The constant is well-chosen at 64 MB — large enough for legitimate payloads (base64-encoded module bundles grow ~33%), small enough to prevent memory exhaustion from a corrupted length prefix. The 8-byte hex prefix allows up to ~4 GB, so without this check a malformed prefix could cause the process to allocate gigabytes. Comment clearly explains the rationale. The constant is exported and tested (`test_max_message_size_is_64mb`).

### src/ftl2/message.py:read_message (size check at line 255-259)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The check is placed at exactly the right location — after the hex prefix is parsed but before the body `reader.read(remaining)` loop. This means an attacker or corrupt stream declaring a 4 GB body will be rejected instantly without allocating memory. The check uses strict `>` (not `>=`), so a message of exactly `MAX_MESSAGE_SIZE` is permitted, which is the correct boundary behavior. The error message includes both the declared size and the limit, which aids debugging. The `ProtocolError` is already caught and re-raised by the outer `except ProtocolError: raise` clause at line 302, so callers will see it correctly.

One minor note: the `send_message` / `send_message_with_id` methods do not enforce the limit on the outgoing side. This is acceptable — the sender controls the data and the purpose of this check is to protect against untrusted/malformed *incoming* data. Adding a send-side check would be defense-in-depth but is not required for correctness.

### tests/test_message.py:test_read_message_exceeds_max_size
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Tests MAX_MESSAGE_SIZE + 1 (the off-by-one boundary). Feeds only the length prefix and EOF — confirming the body is never read. Asserts the correct error substring. Clean and correct.

### tests/test_message.py:test_read_message_at_max_size_is_allowed
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clever boundary test — declares exactly `MAX_MESSAGE_SIZE` but provides only `b"short"` as body, then asserts the failure is "Incomplete message" (not "exceeds maximum"). This proves the size check uses `>` and not `>=`. Well-documented test intent in the docstring and comments. No memory concerns since the read loop will hit EOF after 5 bytes.

### tests/test_message.py:test_read_message_max_hex_prefix_rejected
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Tests the worst-case scenario — `ffffffff` (4,294,967,295 bytes, ~4 GB). Confirms this is rejected by the size check. Important edge case that validates the check protects against the full range of the 8-byte hex prefix format.

### tests/test_message.py:test_max_message_size_is_64mb
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Simple but useful — pins the constant's value, preventing accidental changes. Acts as a regression guard.

### Self-Review
**Limitations:** Could not run the tests to confirm they pass (tool approval issue). Could not inspect whether the gate-side `read_message` call in `ftl_gate/__main__.py` has any special error handling for `ProtocolError` that might need updating, though since `ProtocolError` was already raised for other conditions this is very unlikely to be an issue.

### Feature Requests
- Include the full function body of modified functions in the diff observations, not just the changed hunks — I had to read the full file to verify the check placement was correct relative to the body-read loop
- Show callers of modified functions to verify error handling integration

## gemini [PASS]

### src/ftl2/message.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The addition of a maximum message size check (64 MB) is a solid robustness improvement. It prevents potential memory exhaustion (OOM) crashes that could occur if `reader.readexactly()` attempted to allocate a massive buffer due to a malformed or maliciously crafted length prefix.

### tests/test_message.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The tests are comprehensive and correctly verify the new size limit boundary conditions. They test values just over the limit, exactly at the limit (proving it passes the check and fails on the subsequent read as expected), and the maximum possible 32-bit hex value.

### Self-Review
**Limitations:** None. The changes are self-contained, well-tested, and their intent and correctness are clear from the provided diff.

PR #84, opened 2026-04-11
Fix error handling asymmetry in remote execution paths (GH-75)

### LINKED ISSUE #75: Fix error handling asymmetry between RemoteModuleRunner and LocalModuleRunner
## Summary

`RemoteModuleRunner.run()` raises exceptions on failure while `LocalModuleRunner.run()` returns structured error data. This asymmetry breaks the errors-as-data contract for the AI author, who must handle two different error patterns depending on execution path.

## Problem

The errors-as-data philosophy states: "No failure at any pipeline layer propagates as an exception to user code." This holds for local execution but not for remote:

- **Local path**: `LocalModuleRunner.run()` catches exceptions and returns a dict with `failed=True` — the AI author always gets a dict
- **Remote path**: `RemoteModuleRunner.run()` can raise exceptions (e.g., on connection failures, protocol errors) — the AI author must catch exceptions

For the AI author, this means:
```python
# This might return {"failed": True, ...} OR raise an exception
# depending on whether execution is local or remote
result = await ftl.host.some_module(param="value")
```

The executor layer (`ModuleExecutor`) wraps some of these, but the asymmetry exists at the runner level and can leak through.

## Proposal

Make `RemoteModuleRunner.run()` follow the same contract as `LocalModuleRunner.run()`:
- Catch all exceptions and return structured error dicts with `failed=True`
- Include error classification (transient, permanent, maybe-transient) in the result
- Reserve exception propagation for truly unrecoverable situations (`PolicyDeniedError`, which is intentionally uncaught)

## Context

- The AI author (Claude) treats every result as structured data to reason over. Exceptions break this flow.
- The executor's retry middleware already classifies errors — the runner should return classified results, not raise
- This is listed as a design limitation in the PRD: "RemoteModuleRunner raises while LocalModuleRunner doesn't — asymmetric error handling"

benthomasson:
## Summary

- `_execute_remote_via_gate()` and `_execute_multiplexed()` now wrap their bodies in try/except and return `ExecuteResult(success=False)` instead of propagating exceptions to the caller
- FTL module Error and unexpected gate responses use structured error data (`result_data = {"failed": True, "msg": ...}`) instead of `raise Exception(...)`
- `RuntimeError` for uninitialized runner is preserved as an exception (programming error, not runtime failure)
- Adds 16 tests covering connection failures, protocol errors, broken pipes, gate error responses, and unexpected responses across both serial and multiplexed paths

Closes #75

## Test plan
- [x] All 1562 tests pass (16 new)
- [x] Serial path: gate creation failure returns ExecuteResult
- [x] Serial path: BrokenPipeError returns ExecuteResult
- [x] Serial path: ProtocolError returns ExecuteResult
- [x] Serial path: FTL module Error response returns ExecuteResult (no longer raises)
- [x] Serial path: unexpected response returns ExecuteResult (no longer raises)
- [x] Serial path: RuntimeError for uninitialized runner still raises
- [x] Multiplexed path: protocol error returns ExecuteResult
- [x] Multiplexed path: FTL Error response returns ExecuteResult
- [x] Multiplexed path: BrokenPipeError returns ExecuteResult
- [x] Parametrized test covers ConnectionError, OSError, TimeoutError, ProtocolError, BrokenPipeError, RuntimeError

### DIFF
--- a/src/ftl2/automation/context.py
+++ b/src/ftl2/automation/context.py
@@ -1536,10 +1536,8 @@ class AutomationContext:
 
         except Exception as e:
             logger.exception(f"Remote execution failed for {module_name} on {host.name}")
-            return ExecuteResult(
-                success=False,
-                error=f"Remote execution failed: {e}",
-                output={"failed": True, "msg": f"Remote execution failed: {e}"},
+            return ExecuteResult.from_error(
+                f"Remote execution failed: {e}",
                 module=module_name,
                 host=host.name,
             )
@@ -1676,10 +1674,8 @@ class AutomationContext:
 
         except Exception as e:
             logger.exception(f"Multiplexed execution failed for {module_name} on {host.name}")
-            return ExecuteResult(
-                success=False,
-                error=f"Remote execution failed: {e}",
-                output={"failed": True, "msg": f"Remote execution failed: {e}"},
+            return ExecuteResult.from_error(
+                f"Remote execution failed: {e}",
                 module=module_name,
                 host=host.name,
             )
--- a/src/ftl2/automation/context.py
+++ b/src/ftl2/automation/context.py
@@ -5,6 +5,7 @@ ftl.module_name() syntax for automation scripts.
 """
 
 import asyncio
+import logging
 import os
 import time
 import warnings
@@ -21,6 +22,8 @@ from ftl2.ftl_modules import list_modules, ExecuteResult
 from ftl2.inventory import Inventory, HostGroup, load_inventory, load_localhost
 from ftl2.types import HostConfig, gate_cache_key
 from ftl2.ssh import SSHHost
+
+logger = logging.getLogger(__name__)
 
 
 class OutputMode(Enum):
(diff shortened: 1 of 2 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#84
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **tests/test_context_error_handling.py** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The core fix is correct — the FTL module path previously `raise Exception(...)` on "Error" and unexpected responses while the Ansible module path returned `{"failed": True, ...}`. Changing both to return error data resolves GH-75's asymmetry. However, two issues:

1. **Bare `except Exception` is overly broad.** The outer try/except at line 1387-1545 catches *everything* including `KeyError`, `TypeError`, `AttributeError` — programming bugs that should propagate, not be silently converted to `ExecuteResult(success=False)`. The `logger.exception()` at line 1538 mitigates this somewhat (bugs will appear in logs), but production code will silently "succeed" with an error result when it should crash. Consider narrowing to `(ConnectionError, OSError, BrokenPipeError, ProtocolError, TimeoutError)` or at least excluding `TypeError`/`AttributeError`/`KeyError`.

2. **`ExecuteResult.from_error()` exists but isn't used.** The catch block manually constructs `ExecuteResult(success=False, error=..., output={...}, ...)` — this is exactly what `ExecuteResult.from_error(str(e), module_name, host.name)` does (executor.py:72-98). Using the factory method would be DRY and keep the contract in one place.

3. **`used_ftl` defaults to `True` in the error catch path.** The catch block omits `used_ftl`, so the dataclass default (`True`) applies even for Ansible module failures. Minor but inaccurate.

**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Same asymmetry fix applied correctly — `raise Exception(...)` on "Error" responses replaced with `result_data = {"failed": True, ...}`. Same concerns apply:

1. **Bare `except Exception` too broad** — same issue as the serial path. The multiplexed path has its own try/except *and* is called from within `_execute_remote_via_gate`'s try/except, creating double-catch for programming errors.

2. **Same missing `ExecuteResult.from_error()` usage and `used_ftl` default issue.**

3. The `_execute_multiplexed` path now has defense-in-depth (its own catch + caller's catch), which is fine. But if the inner catch converts a bug to an `ExecuteResult`, the outer catch never fires and the bug is silently swallowed with only a log line.

**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Tests are well-structured and cover the key scenarios (SSH failure, broken pipe, protocol errors, error responses, unexpected responses). Specific concerns:

1. **`test_serial_path_catches_all` parametrizes `RuntimeError`** (line 304) and asserts it's caught — this validates the overly-broad catch behavior. If the catch is narrowed to specific exception types (as recommended), this test would need updating.

2. **Missing: test that `fail_fast` still works end-to-end.** The error-as-data contract works because the caller at context.py:976 converts `success=False` results into `AutomationError` when `fail_fast=True`. No test verifies this integration survives the change.

3. **Missing: test for the "inner raise" in FTL fallback logic.** The inner `except Exception` at lines 1442-1448 re-raises non-import errors. Since those re-raised exceptions are now caught by the outer try/except, the behavior changed — previously they propagated to the caller, now they're silently returned as `ExecuteResult`. No test verifies this changed semantic.

4. **`asyncio.get_event_loop().create_future()`** (lines 222, 249, 281) — deprecated since Python 3.10. Should use `asyncio.get_running_loop().create_future()` or `asyncio.Future()`. Functional in tests but will warn.

### Self-Review
**Limitations:** Could not see the full list of callers of `_execute_remote_via_gate` and `_execute_multiplexed` beyond what grep showed. Could not verify whether any caller depends on exceptions propagating (e.g., for retry logic or gate cleanup). Did not see the existing test suite to check for overlapping coverage or tests that might break from the behavioral change.

### Feature Requests
- Include callers of modified methods and show how they handle the return value, so reviewers can assess whether error-swallowing is safe end-to-end
- Show any existing tests for the same functions so duplicate/conflicting coverage can be flagged
- When a dataclass has a `from_error()`/`from_*()` factory method, flag catch blocks that manually construct the same shape

## gemini [PASS]

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The changes gracefully wrap remote execution logic in `try/except` blocks, catching connection, protocol, and execution exceptions and returning them as an `ExecuteResult` with `success=False`. This correctly enforces an "errors-as-data" pattern instead of raising exceptions. Logging is used appropriately, and uninitialized component errors (e.g. `RuntimeError`) are purposely kept outside the block to ensure programming errors are still surfaced immediately.

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The tests are comprehensive, mocking various networking, protocol, and internal exceptions (like `BrokenPipeError` and `ProtocolError`) for both the serial and multiplexed execution paths. They effectively assert that an `ExecuteResult` object is returned representing a failure rather than bubbling up exceptions.

### Self-Review
**Limitations:** None. The diff and context are complete and self-contained.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#84
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **tests/test_context_error_handling.py** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

**The good:** The core change is sound — wrapping the method body in `try/except Exception` and returning `ExecuteResult.from_error()` instead of raising aligns with the errors-as-data contract. The `RuntimeError` guard for uninitialized runner correctly sits *outside* the try block (line 1384), so programming errors still propagate. The caller `run_on()` (line 1311) already expects `ExecuteResult` return values and properly checks `result.success`. The `fail_fast` path (line 976) works correctly since it checks `result.success` on the returned value.

**Behavioral change in FTL Error path (line 1434-1435):** The original code did `raise Exception(response[1].get("message", ...))` for "Error" responses from the gate. The new code sets `result_data = {"failed": True, ...}` instead. This is actually a *double* fix: the old `raise` would be caught by the inner `except Exception` (line 1442), where `str(e)` would not contain "No module named" or "ImportError", so it would `raise` again — which would then propagate uncaught (pre-PR). Now it's handled cleanly. Same fix for the `else` (unexpected response) branch at line 1436-1437. This is correct.

**Concern — `logger.exception()` with f-string (line 1538):** Uses `logger.exception(f"Remote execution failed for {module_name} on {host.name}")`. This works but is a minor logging anti-pattern — the f-string is evaluated even if the log level is disabled. Should use `logger.exception("Remote execution failed for %s on %s", module_name, host.name)`. Not a blocker, but `exception()` always logs at ERROR level so it's always evaluated anyway. Minor point.

**Concern — bare `except Exception` is very broad:** This catches *everything* including `KeyboardInterrupt` subclasses (no, actually `KeyboardInterrupt` inherits from `BaseException`, not `Exception`, so it won't be caught). But it does catch things like `AttributeError`, `TypeError`, `KeyError` — bugs in the method itself would be silently swallowed and returned as "Remote execution failed" results. This makes debugging harder. A more targeted approach would catch `(OSError, BrokenPipeError, ConnectionError, ProtocolError, TimeoutError)` explicitly. However, since `logger.exception()` logs the full traceback, the information isn't lost — just harder to surface. This is a tradeoff.

**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Same pattern as `_execute_remote_via_gate`. The try/except wraps the full method body and returns `ExecuteResult.from_error()`. Same behavioral fix: "Error" responses from FTL modules now produce `result_data` dicts instead of raising (old line: `raise Exception(resp_data.get("message", ...))`). This is correct and consistent.

**Same concern about broad `except Exception`** — bugs in the multiplexed path (e.g., `AttributeError` from a `None` gate field) would be silently converted to "Remote execution failed" results rather than crashing with a useful traceback. The `logger.exception()` mitigates this but only if someone is watching the logs.

**Same f-string logging concern** at line 1683.

**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

**Good coverage:** Tests cover gate creation failure, BrokenPipeError, ProtocolError, Error response, unexpected response, RuntimeError still propagating, output dict structure, and multiplexed path equivalents. The parametrized `test_serial_path_catches_all` is a nice touch.

**Patch location correctness:** Tests patch `ftl2.ftl_modules.executor.is_ftl_module` — this works because the function-local `from ftl2.ftl_modules.executor import is_ftl_module` re-reads the module attribute each time the function is called, picking up the mock.

**Concern — `ProtocolError` imported in test but never imported in context.py:** The test file imports `ProtocolError` from `ftl2.message` (line 8 of test) and uses it in `test_protocol_error_returns_result` and the parametrized test. The *production code* (`context.py`) does not import or specifically handle `ProtocolError` — it catches it via the broad `except Exception`. This is fine functionally, but the test creates an illusion that `ProtocolError` is a specifically-handled case when it's actually just caught by the generic handler. Not a bug, but slightly misleading.

**Concern — `test_runtime_error_still_raised` has a subtle issue:** The test sets `ctx._remote_runner = None` and expects `RuntimeError`. But the test doesn't set up `ctx._gate_locks` — that's fine because the RuntimeError fires before any lock access. However, this test is somewhat fragile: if someone moves the RuntimeError guard inside the try block, the test would still pass (because RuntimeError is a subclass of Exception and would be caught, returning an ExecuteResult), silently breaking the intended contract. Consider asserting that the error message is NOT wrapped in "Remote execution failed:" to detect this regression.

**Missing test:** No test for the case where `_execute_multiplexed` is called from within `_execute_remote_via_gate` and both have try/except — verifying no double-wrapping of errors.

**Missing test for parametrized multiplexed path:** `TestErrorDataContract` only tests the serial path. A symmetric parametrized test for the multiplexed path would strengthen coverage.

### Self-Review
**Limitations:** - Could not see the full `run_on` method or `execute` method that calls `_execute_remote_via_gate` for the local execution path to verify complete symmetry
- Did not verify whether any integration tests or end-to-end tests exist that might break from the behavioral change (Error responses no longer raising)
- Could not check git history for GH-75 issue description to understand the original bug report

### Feature Requests
- Include the issue/PR description (GH-75) when referenced in commit messages so reviewers can verify issue compliance
- Show callers of modified methods automatically to verify integration impact

## gemini [PASS]

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

The modifications correctly replace the previous `raise Exception(...)` calls with inline assignments to `result_data` representing the error, and wrap the entire execution blocks of `_execute_remote_via_gate` and `_execute_multiplexed` in `try...except` to catch runtime and network failures. The caught exceptions are elegantly returned as `ExecuteResult` instances, respecting the "errors-as-data" contract.

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

The new test file provides comprehensive coverage for both serial and multiplexed paths, specifically validating that various exceptions (e.g., ConnectionError, ProtocolError, BrokenPipeError) and error responses are properly caught and converted into `ExecuteResult` objects, rather than bubbling up. Mocks are configured appropriately for async methods.

### Self-Review
**Limitations:** I did not have the source file for `ftl2/ftl_modules/executor.py` so I could not verify the exact implementation of `ExecuteResult.from_error`, but its usage perfectly mirrors the assertions in the provided test file.

PR #95, opened 2026-04-11
Add periodic Hello keepalive and send_and_wait timeout for hung gate detection

### LINKED ISSUE #69: Add periodic Hello keepalive and send_and_wait timeout for hung gate detection

benthomasson:
Closes #69

### DIFF
--- a/src/ftl2/exceptions.py
+++ b/src/ftl2/exceptions.py
@@ -130,6 +130,8 @@ class ErrorTypes:
     PERMISSION_DENIED = "PermissionDenied"
     INVENTORY_ERROR = "InventoryError"
     GATE_ERROR = "GateError"
+    GATE_TIMEOUT = "GateTimeout"
+    GATE_UNRESPONSIVE = "GateUnresponsive"
     UNKNOWN = "Unknown"
 
 
@@ -351,6 +353,49 @@ class InventoryError(FTL2Error):
             message=message,
         )
         super().__init__(message, context)
+
+
+class GateRequestTimeoutError(ModuleExecutionError):
+    """Raised when a gate request exceeds the per-request timeout.
+
+    Extends ModuleExecutionError so callers catching the parent type
+    still handle it. Indicates an individual request hung — the gate
+    may still be healthy for other requests.
+    """
+
+    def __init__(self, message: str, host: str = "", module: str = ""):
+        super().__init__(
+            message, host=host, module=module,
+            error_type=ErrorTypes.GATE_TIMEOUT,
+        )
+
+
+class GateHandshakeTimeoutError(FTL2ConnectionError):
+    """Raised when the initial Hello handshake with a gate times out.
+
+    Indicates the gate process failed to start or respond.
+    """
+
+    def __init__(self, message: str, host: str = ""):
+        super().__init__(
+            message, host=host,
+            error_type=ErrorTypes.GATE_TIMEOUT,
+        )
+
+
+class GateUnresponsiveError(FTL2ConnectionError):
+    """Raised when a gate fails periodic keepalive health checks.
+
+    Indicates the gate process is wedged (deadlock, hang, resource
+    exhaustion). All pending futures on the connection are failed
+    with this error.
+    """
+
+    def __init__(self, message: str, host: str = ""):
+        super().__init__(
+            message, host=host,
+            error_type=ErrorTypes.GATE_UNRESPONSIVE,
+        )
 
 
 class ExcludedModuleError(FTL2Error):
--- a/src/ftl2/ftl_gate/__main__.py
+++ b/src/ftl2/ftl_gate/__main__.py
@@ -1433,6 +1433,13 @@ async def main_multiplexed(reader, writer, protocol, watcher, monitor, gate_hash
         """Handle a single request and send response with same msg_id."""
         nonlocal gate_policy, gate_environment, gate_host
         try:
+            if msg_type == "Hello":
+                await protocol.send_message_with_id(
+                    writer, "Hello", {"gate_hash": gate_hash},
+                    msg_id, write_lock=write_lock,
+                )
+                return
+
             if msg_type == "Module":
                 if not isinstance(data, dict):
                     await protocol.send_message_with_id(
--- a/src/ftl2/runners.py
+++ b/src/ftl2/runners.py
@@ -27,6 +27,9 @@ from .exceptions import (
     ErrorContext,
     ErrorTypes,
     FTL2Error,
+    GateHandshakeTimeoutError,
+    GateRequestTimeoutError,
+    GateUnresponsiveError,
     ModuleExecutionError,
     get_suggestions,
 )
@@ -106,9 +109,11 @@ class Gate:
     temp_dir: str
     interpreter: str = "python3"
     multiplexed: bool = False
+    healthy: bool = True
     _msg_counter: int = field(default=0, repr=False)
     _pending: dict[int, asyncio.Future] = field(default_factory=dict, repr=False)
     _reader_task: asyncio.Task | None = field(default=None, repr=False)
+    _keepalive_task: asyncio.Task | None = field(default=None, repr=False)
     _write_lock: asyncio.Lock = field(default_factory=asyncio.Lock, repr=False)
 
     def next_msg_id(self) -> int:
@@ -548,6 +553,12 @@ class RemoteModuleRunner(ModuleRunner):
         >>> result = await runner.run(host, context)
     """
 
+    # Timeout constants (seconds)
+    REQUEST_TIMEOUT: float = 300.0       # Per-request timeout for _send_and_wait
+    HANDSHAKE_TIMEOUT: float = 30.0      # Hello handshake timeout during gate startup
+    KEEPALIVE_INTERVAL: float = 30.0     # Seconds between keepalive pings
+    KEEPALIVE_TIMEOUT: float = 15.0      # Max wait for keepalive response
+
     def __init__(self) -> None:
         """Initialize the remote runner with empty gate cache."""
         self.gate_cache: dict[str, Gate] = {}
@@ -691,7 +702,11 @@ class RemoteModuleRunner(ModuleRunner):
         # Check cache first
         if cache_key in self.gate_cache:
             gate = self.gate_cache[cache_key]
-            if gate.interpreter == interpreter:
+            if not gate.healthy:
+                logger.info(f"Evicting unhealthy gate for {cache_key}")
+                del self.gate_cache[cache_key]
+                await self._close_gate(gate)
+            elif gate.interpreter == interpreter:
                 logger.debug(f"Reusing cached gate for {cache_key}")
                 # Multiplexed gates stay in cache (shared access).
                 # Serial gates are popped (exclusive access).
@@ -705,7 +720,7 @@ class RemoteModuleRunner(ModuleRunner):
 
         # Create new gate connection
         logger.info(f"Creating new gate for {cache_key}")
-        return await self._connect_gate(ssh_host, ssh_port, ssh_user, ssh_password, ssh_key_file, interpreter, context, host_name=host_name)
+        return await self._connect_gate(ssh_host, ssh_port, ssh_user, ssh_password, ssh_key_file, interpreter, context, host_name=host_name, cache_key=cache_key)
 
     async def _connect_gate(
         self,
@@ -721,6 +736,7 @@ class RemoteModuleRunner(ModuleRunner):
         become: "BecomeConfig | None" = None,
         event_callback: Any = None,
         host_name: str = "localhost",
+        cache_key: str = "",
     ) -> Gate:
         """Establish SSH connection and create gate.
 
@@ -806,6 +822,9 @@ class RemoteModuleRunner(ModuleRunner):
                 if supports_multiplex:
                     gate._reader_task = asyncio.create_task(
                         _gate_reader_loop(gate, self.protocol, event_callback)
+                    )
+                    gate._keepalive_task = asyncio.create_task(
+                        self._keepalive_loop(gate, cache_key)
                     )
                     logger.info(f"Multiplexed gate for {ssh_host}:{ssh_port}")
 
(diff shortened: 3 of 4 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#95
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/exceptions.py — New error types and ErrorTypes constants
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Three well-designed exception classes with clean hierarchy. `GateRequestTimeoutError` extends `ModuleExecutionError` (correct — a request-level failure, callers catching `ModuleExecutionError` still handle it). `GateHandshakeTimeoutError` and `GateUnresponsiveError` extend `FTL2ConnectionError` (correct — these are connection-level failures). The `ErrorTypes` constants `GATE_TIMEOUT` and `GATE_UNRESPONSIVE` are used consistently. The `TestExceptionHierarchy` test class validates these relationships.

### src/ftl2/ftl_gate/__main__.py:handle_request — Hello handler in multiplexed mode
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The gate-side Hello handler in `handle_request` (line ~1188) responds with `{"gate_hash": gate_hash}` but ignores the incoming `data` dict. Compare with the serial-mode Hello handler (line 942-959) which echoes back `dict(data)` and merges `gate_hash` into it. This means the keepalive Hello response omits any data the controller sent — not a bug for keepalive (controller sends `{}`) but a subtle behavioral difference vs the serial path. More importantly, this gate-side change is **not directly tested** in `test_gate_timeouts.py` since those tests mock the protocol layer. Consider adding an integration-level test or at least a note that the multiplexed Hello handler is intentionally simpler.

### src/ftl2/runners.py:Gate dataclass — New `healthy` and `_keepalive_task` fields
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

`healthy: bool = True` and `_keepalive_task` field additions are clean. The `healthy` field is checked in `_get_or_create_gate`, set in `_close_gate` and `_keepalive_loop`. Tests in `TestGateDataclass` verify defaults. Using a plain bool is adequate — the gate is single-owner (one runner), so no thread-safety concern.

### src/ftl2/runners.py:_send_and_wait — Per-request timeout
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

`asyncio.wait_for(future, timeout=timeout)` correctly wraps the future. On timeout, the pending future is cleaned up via `gate._pending.pop(msg_id, None)` — this prevents a stale future from lingering. The `from None` suppresses the chained `TimeoutError`, which is clean since we raise a more informative `GateRequestTimeoutError`. The default 300s matches `_drain_gate`'s timeout. All existing callers (`_execute_through_gate`, `_execute_with_upload`, `run_ftl_module`) use the default timeout path. Tests cover timeout, cleanup, concurrent timeouts, and success cases.

### src/ftl2/runners.py:_open_gate — Handshake timeout
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `asyncio.wait_for` wrapping of `read_message` with `HANDSHAKE_TIMEOUT` (30s default) correctly prevents indefinite hangs during gate startup. The stderr collection has its own 5s timeout (preventing double-hang), and both timeout paths are tested. The try/except around stderr read catches `Exception` broadly, which is appropriate for diagnostic-only code.

### src/ftl2/runners.py:_keepalive_loop — Periodic health checking
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The keepalive loop has solid test coverage and correct happy-path logic. However, there is a **race condition concern**: when keepalive fails, the loop calls `gate.gate_process.stdin.close()` directly, but the `_gate_reader_loop` is still running and may also be trying to fail pending futures (lines 142-147 in `_gate_reader_loop`). The keepalive loop clears `_pending` after setting exceptions on futures, while the reader loop also iterates `_pending` on EOF — if `stdin.close()` triggers EOF on stdout (common with SSH channels), the reader loop will iterate an already-cleared `_pending` dict, which is benign. But if timing is different and the reader loop sees EOF *before* keepalive clears `_pending`, futures could receive exceptions from **both** paths. This is safe (`future.done()` guards in both) but worth noting. Also, the keepalive loop does **not** cancel the `_reader_task` after detecting failure — it relies on the reader loop noticing the closed stdin eventually, creating a brief window where the reader task is still running on a dead gate.

### src/ftl2/runners.py:_get_or_create_gate — Unhealthy gate eviction
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean eviction logic: check `gate.healthy`, delete from cache, call `_close_gate`. The `elif` preserving the existing interpreter check is correct — unhealthy gates are prioritized over interpreter mismatches. `_close_gate` is awaited before creating a new gate, ensuring resources are released. Test `TestUnhealthyGateEviction` verifies the full flow.

### src/ftl2/runners.py:_close_gate — Keepalive cleanup
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The diff shows `gate.healthy = False` at the top and keepalive task cancellation after reader task cancellation. Looking at the **current** `_close_gate` (lines 1308-1344), the diff inserts the `healthy = False` before reader cancellation and the keepalive cancellation block between reader cancellation and the pending-futures block. This ordering is correct: marking unhealthy first lets the keepalive loop exit its `while gate.healthy` check promptly, then explicit cancellation ensures cleanup. However, based on the diff, `_close_gate` cancels the **reader task first, then keepalive task**, then fails pending futures. But if keepalive is mid-flight in `_send_and_wait` (which creates its own pending future), the `_close_gate`'s "fail pending futures" block will catch it. This is fine but the ordering could be slightly cleaner if keepalive were cancelled before the reader (to avoid the keepalive adding new pending entries while we're failing them). In practice, `gate.healthy = False` plus the sleep in keepalive makes this race window negligible.

### src/ftl2/runners.py:_connect_gate — Keepalive task creation
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Keepalive task is created only for multiplexed gates, right after the reader task — correct placement. The `cache_key` parameter addition to `_connect_gate` is clean. However, the keepalive task creation path itself isn't directly unit-tested (it's tested through the keepalive loop tests), and the new `cache_key` parameter on `_connect_gate` should be verified against all callers.

### src/ftl2/runners.py — Timeout constants
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

`REQUEST_TIMEOUT=300s`, `HANDSHAKE_TIMEOUT=30s`, `KEEPALIVE_INTERVAL=30s`, `KEEPALIVE_TIMEOUT=15s`. Values are sensible. `KEEPALIVE_TIMEOUT < KEEPALIVE_INTERVAL` is enforced by test. Constants are class-level attributes, easily overridden in tests (which the tests do). One note: `REQUEST_TIMEOUT` matches the existing `_drain_gate` timeout of 300s, which is a good consistency signal.

### tests/test_gate_timeouts.py — Test suite
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test suite with 478 lines covering all new functionality: `_send_and_wait` timeout (6 tests), handshake timeout (3 tests), keepalive loop (5 tests), unhealthy gate eviction (1 test), `_close_gate` lifecycle (3 tests), Gate dataclass fields (2 tests), timeout constants (2 tests), exception hierarchy (5 tests). Tests use short timeouts (0.02-0.05s) for fast execution. The `_make_gate` helper is clean. Tests properly use `asyncio.CancelledError` and `asyncio.wait_for` for timeout scenarios. One minor note: `test_keepalive_connection_error_exits_cleanly` patches `_send_and_wait` directly rather than the protocol, which means it doesn't test the actual exception propagation path from the protocol layer, but this is acceptable for unit testing the loop logic.

### `_drain_gate` — Existing code not updated
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

`_drain_gate` (line 1346) still uses raw `asyncio.wait_for(future, timeout=...)` and `gate.create_future(msg_id)` directly instead of going through `_send_and_wait`. This means drain operations don't benefit from the new `GateRequestTimeoutError` wrapping — they'll raise raw `asyncio.TimeoutError` instead. This isn't a bug (drain has its own timeout semantics with `timeout_seconds + 30`), but it's an inconsistency. More importantly, if `_drain_gate` times out, it doesn't clean up the future from `gate._pending`, unlike the updated `_send_and_wait` which does `gate._pending.pop(msg_id, None)`.

### Self-Review
**Limitations:** I could not see the full `_close_gate` method *after* the diff is applied (only the pre-patch version plus the diff hunks). I reconstructed the post-patch version mentally but may have the insertion ordering slightly wrong. I also could not verify whether other callers of `_connect_gate` (e.g., subsystem registration) pass the new `cache_key` parameter or rely on the default empty string. The `_gate_reader_loop` is a standalone function (not a method), so its interaction with keepalive cleanup during concurrent failure is inferred from code structure, not tested.

### Feature Requests
- Show the full post-patch version of modified functions, not just pre-patch plus diff hunks, to make insertion-order verification easier
- Include a caller graph for modified method signatures to verify all call sites pass new parameters
- Flag when existing code with similar patterns (like `_drain_gate` using raw `wait_for`) isn't updated to use the new abstraction

## gemini [PASS]

### src/ftl2/exceptions.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Exception classes are correctly defined and inherit from the appropriate base classes (`ModuleExecutionError` for request timeouts and `FTL2ConnectionError` for handshake/keepalive timeouts), which ensures correct high-level error handling.

### src/ftl2/ftl_gate/__main__.py:handle_request
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The remote gate correctly intercepts "Hello" messages after initialization and sends back a response with the same `msg_id`, successfully acting as an echo/keepalive mechanism.

### src/ftl2/runners.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The connection timeouts and keepalive loops are robustly implemented. The code properly uses `asyncio.wait_for`, handles `TimeoutError` cleanly, fails pending futures with appropriate exceptions upon connection death, and securely evicts unhealthy gates from the cache without race conditions. Safe execution is ensured by falling back to `stderr` reading during handshake timeouts with its own safety timeout.

### tests/test_gate_timeouts.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Excellent comprehensive unit tests. The file covers the new timeout paths, exceptions, future cleanup, keepalive lifecycle, and cache eviction logic thoroughly by mocking out the SSH layer.

### Self-Review
**Limitations:** None. The diff and context provided everything necessary to thoroughly verify the concurrency and timeout implementation within the asyncio event loop context.

PR #96, opened 2026-04-11
Add gate-side policy enforcement for defense-in-depth

### LINKED ISSUE #70: Add gate-side policy enforcement for defense-in-depth

benthomasson:
Closes #70

### DIFF
--- a/src/ftl2/automation/context.py
+++ b/src/ftl2/automation/context.py
@@ -1503,6 +1503,8 @@ class AutomationContext:
                             # Gate wraps module output in {"result": ...} — unwrap it
                             if "result" in result_data and isinstance(result_data["result"], dict):
                                 result_data = result_data["result"]
+                        elif response is not None and response[0] == "PolicyDenied":
+                            result_data = {"failed": True, "msg": f"Gate policy denied: {response[1].get('reason', '')}"}
                         elif response is not None and response[0] == "Error":
                             result_data = {"failed": True, "msg": response[1].get("message", "Unknown FTL module error")}
                         else:
@@ -1586,6 +1588,8 @@ class AutomationContext:
                                     "stdout": stdout,
                                     "stderr": stderr,
                                 }
+                        elif msg_type == "PolicyDenied":
+                            result_data = {"failed": True, "msg": f"Gate policy denied: {data.get('reason', '')}"}
                         elif msg_type == "Error":
                             result_data = {"failed": True, "msg": data.get("message", "Unknown error")}
                         else:
@@ -1663,6 +1667,8 @@ class AutomationContext:
                         result_data = dict(resp_data)
                         if "result" in result_data and isinstance(result_data["result"], dict):
                             result_data = result_data["result"]
+                    elif resp_type == "PolicyDenied":
+                        result_data = {"failed": True, "msg": f"Gate policy denied: {resp_data.get('reason', '')}"}
                     elif resp_type == "Error":
                         result_data = {"failed": True, "msg": resp_data.get("message", "Unknown FTL module error")}
                     else:
@@ -1728,6 +1734,8 @@ class AutomationContext:
                             "stdout": stdout,
                             "stderr": stderr,
                         }
+                elif resp_type == "PolicyDenied":
+                    result_data = {"failed": True, "msg": f"Gate policy denied: {resp_data.get('reason', '')}"}
                 elif resp_type == "Error":
                     result_data = {"failed": True, "msg": resp_data.get("message", "Unknown error")}
                 else:
@@ -1820,6 +1828,7 @@ class AutomationContext:
             register_subsystem=self._gate_subsystem,
             become=become,
             event_callback=_event_cb,
+            host_name=host.name,
         )
 
     async def _get_ssh_connection(self, host: HostConfig) -> SSHHost:
@@ -1853,6 +1862,8 @@ class AutomationContext:
 
         self._remote_runner = RemoteModuleRunner()
         self._remote_runner.gate_builder = GateBuilder()
+        self._remote_runner.policy_wire = self._policy.to_wire()
+        self._remote_runner.environment = self._environment
         self._bundle_cache = BundleCache()
         self._resolve_gate_modules()
         self._start_time = time.time()
--- a/src/ftl2/ftl_gate/__main__.py
+++ b/src/ftl2/ftl_gate/__main__.py
@@ -52,7 +52,46 @@ try:
 except ImportError:
     HAS_FTL_GATE = False
 
+# Try to import policy engine (bundled into gate .pyz for gate-side enforcement)
+try:
+    from ftl2.policy import Policy
+    HAS_POLICY = True
+except ImportError:
+    HAS_POLICY = False
+
 logger = logging.getLogger("ftl_gate")
+
+
+def _check_gate_policy(
+    policy: "Policy | None",
+    module_name: str,
+    module_args: dict[str, Any],
+    environment: str = "",
+    host: str = "localhost",
+) -> tuple[bool, dict[str, Any] | None]:
+    """Evaluate gate-side policy before module execution.
+
+    Args:
+        policy: Active policy instance, or None if no policy is set.
+        module_name: Name of the module to execute.
+        module_args: Module arguments dict.
+        environment: Environment label for environment-scoped rules.
+        host: Logical host name for host-scoped rules.
+
+    Returns:
+        Tuple of (permitted, denial_data). If permitted is False,
+        denial_data contains the structured PolicyDenied response payload.
+    """
+    if policy is None or not policy.rules:
+        return True, None
+    result = policy.evaluate(module_name, module_args, host=host, environment=environment)
+    if result.permitted:
+        return True, None
+    return False, {
+        "module": module_name,
+        "reason": result.reason,
+        "rule": result.rule.to_dict() if result.rule else None,
+    }
 
 
 class ModuleNotFoundError(Exception):
@@ -919,6 +958,11 @@ async def main(args: list[str]) -> int | None:
     watcher = FileWatcher(protocol, writer)
     monitor = SystemMonitor(protocol, writer)
 
+    # Gate-side policy enforcement state
+    gate_policy: "Policy | None" = None
+    gate_environment: str = ""
+    gate_host: str = "localhost"
+
     # Message processing loop
     while True:
         try:
@@ -944,6 +988,15 @@ async def main(args: list[str]) -> int | None:
                 response_data = dict(data) if isinstance(data, dict) else {}
                 response_data["gate_hash"] = gate_hash
 
+                # Extract policy rules and host from Hello data
+                if isinstance(data, dict) and HAS_POLICY:
+                    policy_rules = data.get("policy_rules")
+                    if policy_rules:
+                        gate_policy = Policy.from_wire(policy_rules)
+                        gate_environment = data.get("environment", "")
+                        gate_host = data.get("host", "localhost")
+                        logger.info(f"Policy loaded: {len(gate_policy.rules)} rules, environment={gate_environment!r}, host={gate_host!r}")
+
                 # Check for multiplexing capability
                 capabilities = data.get("capabilities", []) if isinstance(data, dict) else []
                 if "multiplex" in capabilities:
@@ -954,7 +1007,7 @@ async def main(args: list[str]) -> int | None:
                     else:
                         await protocol.send_message(writer, "Hello", response_data)
                     logger.info("Entering multiplexed mode")
-                    return await main_multiplexed(reader, writer, protocol, watcher, monitor, gate_hash)
+                    return await main_multiplexed(reader, writer, protocol, watcher, monitor, gate_hash, gate_policy, gate_environment, gate_host)
                 else:
                     await protocol.send_message(writer, "Hello", response_data)
 
@@ -965,6 +1018,15 @@ async def main(args: list[str]) -> int | None:
                     await protocol.send_message(
                         writer, "Error", {"message": "Invalid Module data"}
                     )
+                    continue
+
+                # Gate-side policy check
+                permitted, denial = _check_gate_policy(
+                    gate_policy, data.get("module_name", ""), data.get("module_args", {}), gate_environment, gate_host
+                )
+                if not permitted:
+                    logger.info(f"Policy denied Module: {denial}")
+                    await protocol.send_message(writer, "PolicyDenied", denial)
                     continue
 
                 try:
@@ -1001,6 +1063,15 @@ async def main(args: list[str]) -> int | None:
                     await protocol.send_message(
                         writer, "Error", {"message": "Invalid FTLModule data"}
                     )
+                    continue
+
+                # Gate-side policy check
+                permitted, denial = _check_gate_policy(
+                    gate_policy, data.get("module_name", ""), data.get("module_args", {}), gate_environment, gate_host
+                )
+                if not permitted:
+                    logger.info(f"Policy denied FTLModule: {denial}")
+                    await protocol.send_message(writer, "PolicyDenied", denial)
                     continue
 
                 try:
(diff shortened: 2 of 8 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#96
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## Disagreements

- **tests/test_gate_policy_enforcement.py** [LOW]: claude: CONCERN, gemini: PASS

## claude [BLOCK]

### src/ftl2/policy.py:PolicyRule.to_dict
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Simple serialization of dataclass fields to dict. Correctly returns all three fields (decision, match, reason). Used by `_check_gate_policy` to include the matching rule in denial data. Well tested in `TestPolicyRuleToDict`.

### src/ftl2/policy.py:Policy.to_wire
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Delegates to `rule.to_dict()` for each rule. Used in `context.py:__aenter__` to set `runner.policy_wire`. Well tested including JSON round-trip and order preservation.

### src/ftl2/policy.py:Policy.from_wire
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Uses `PolicyRule(**entry)` which will raise `TypeError` on malformed dicts (extra keys, missing keys). Since the gate receives this from an SSH channel, a corrupted or malicious wire payload would crash the gate process rather than returning an error. Consider wrapping the construction in a try/except or validating keys. This is a moderate concern since the SSH channel is trusted, but defense-in-depth matters for a policy enforcement system.

### src/ftl2/ftl_gate/__main__.py:_check_gate_policy
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean helper function. Returns early for `None` policy or empty rules. Correctly calls `policy.evaluate()` with all four parameters. Returns structured denial data with module, reason, and serialized rule. Thoroughly tested including host scoping, environment scoping, module equivalence, and param matching.

### src/ftl2/ftl_gate/__main__.py:main (Hello policy extraction)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Correctly extracts `policy_rules` from Hello data and creates a `Policy` instance. Gracefully skips when `HAS_POLICY` is False (import failed). Extracts `environment` and `host` with sensible defaults. The `HAS_POLICY` guard is well-placed. Tests verify the import flag and Hello data format but don't exercise the actual gate main loop end-to-end.

### src/ftl2/ftl_gate/__main__.py:main (Module/FTLModule policy checks)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Policy check is correctly placed before `execute_module` and `execute_ftl_module` calls in both non-multiplexed paths. Uses `continue` to skip execution on denial, sends `PolicyDenied` response. Duplicated in 4 places (Module non-mux, FTLModule non-mux, Module mux, FTLModule mux) — correct but verbose. No integration test exercises the actual message flow.

### src/ftl2/ftl_gate/__main__.py:main_multiplexed (SetPolicy handler)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

The `SetPolicy` handler in `main_multiplexed` uses `nonlocal gate_policy, gate_environment, gate_host` to update shared state from concurrent tasks. Since `main_multiplexed` handles requests concurrently via `asyncio.create_task`, a `SetPolicy` running concurrently with a `Module` task creates a race condition: a Module could read `gate_policy` while `SetPolicy` is partway through updating `gate_policy` + `gate_environment` + `gate_host`. In practice, Python's GIL and asyncio's cooperative scheduling mean individual assignments are atomic, but the three assignments are NOT atomic as a group — a Module handler could see the new policy but the old environment. This is a real concern.

### src/ftl2/message.py (new message types)
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

**Naming conflict**: The diff adds `SetPolicy`/`SetPolicyResult` to MESSAGE_TYPES, but the working tree already has `UpdatePolicy`/`UpdatePolicyResult` (added by a different change). These appear to serve the same purpose ("push policy rules to a running gate"). The gate code (`__main__.py`) handles `SetPolicy` but nothing handles `UpdatePolicy`. Additionally, the test asserts `len(protocol.MESSAGE_TYPES) == 25` but after merging with the working tree (which already has 27 types: 22 base + UpdatePolicy + UpdatePolicyResult + Goodbye + GateDrain + GateDrainResult), the count would be 30. The test will fail when applied to the actual working tree. **Resolution needed**: either rename `UpdatePolicy` to `SetPolicy` everywhere, or vice versa, and fix the count assertion.

### src/ftl2/gate.py:_copy_policy_module
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Raises `GateError` on failure, which is inconsistent with `_copy_exceptions_module` (which only logs a warning) and `_copy_ftl_module_exceptions` (also logs a warning). If `policy.py` is missing, the gate build fails hard. This may be intentional (policy is required), but the caller `_build_new_gate` calls both methods sequentially — if `_copy_policy_module` raises, `_copy_exceptions_module` never runs, leaving a partially built gate. Consider whether a warning + graceful degradation (like `_copy_exceptions_module`) is more appropriate, since the gate already handles `HAS_POLICY = False` gracefully.

### src/ftl2/gate.py:_copy_exceptions_module
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Correctly copies `exceptions.py` as a dependency of `policy.py`. Uses warning-level logging on failure (graceful degradation). Existence tests verify the method is present and that the source file exists.

### src/ftl2/gate.py:compute_hash
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly adds `policy.py` and `exceptions.py` to the hash computation, ensuring that changes to the policy engine invalidate the gate cache. The `if source_file.exists()` guard handles the case where these files might not exist.

### src/ftl2/runners.py:RemoteModuleRunner (policy_wire/environment attrs)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

New `policy_wire` and `environment` attributes with sensible defaults (empty list, empty string). Set in `context.py:__aenter__` from `self._policy.to_wire()` and `self._environment`. Tests verify attribute existence and defaults.

### src/ftl2/runners.py:_open_gate (Hello with policy data)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Conditionally includes `policy_rules`, `environment`, and `host` in Hello data only when `policy_wire` is non-empty. This avoids unnecessary payload when no policy is configured. Clean implementation.

### src/ftl2/runners.py:_execute_through_gate (PolicyDenied handling)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Correctly raises `PolicyDeniedError` when gate returns `PolicyDenied`. Placed before the generic `Error` handler. Error message includes both module name and reason from the denial data. Consistent across `_execute_through_gate`, `_execute_with_upload`, and `run_ftl_module`.

### src/ftl2/runners.py:host_name parameter threading
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

The `host_name` parameter is threaded through `run() -> _get_or_create_gate() -> _connect_gate() -> _open_gate()` in `runners.py`. However, `context.py:_get_or_create_gate` (line 1874) also calls `_connect_gate` and does **NOT** pass `host_name`. This means when gates are created through the automation context path (which is the primary usage path for `_execute_remote_via_gate` and `_execute_multiplexed`), the host defaults to `"localhost"` and host-scoped policy rules won't be enforced on the gate side. This is a significant integration gap.

### src/ftl2/automation/context.py:__aenter__ (policy_wire setup)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Correctly sets `policy_wire` and `environment` on the runner during context entry. This ensures all gates created through this context carry the policy.

### src/ftl2/automation/context.py:_execute_remote_via_gate (PolicyDenied handling)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Correctly handles `PolicyDenied` responses in all four code paths: non-mux Module, non-mux FTLModule, multiplexed Module, multiplexed FTLModule. Returns `result_data` with `failed: True` and descriptive message rather than raising, consistent with how `Error` responses are handled. This allows the `ExecuteResult` to be constructed normally downstream.

### tests/test_gate_policy_enforcement.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** PARTIAL

Comprehensive test suite covering serialization, round-trips, `_check_gate_policy`, message types, builder methods, runner attributes, and edge cases. 615 lines with good coverage of the policy enforcement logic. **Issues**: (1) `test_message_types_defined` in `test_message.py` asserts count == 25 which won't match the working tree; (2) No integration tests that exercise the actual gate message loop with policy enforcement — all tests are unit-level; (3) The `test_hash_changes_with_policy_content` test doesn't actually test that the hash changes — it only verifies that two identical configs produce the same hash (the name is misleading).

### tests/test_message.py (count assertion)
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** PARTIAL

The diff changes the assertion from `== 22` to `== 25`, but the working tree already has 27 message types (22 base + 5 from other uncommitted changes). When both sets of changes are applied, there will be 30 types, and the assertion will fail. This needs reconciliation with the other in-flight changes.

### Self-Review
**Limitations:** - Could not run the tests to verify the count assertion conflict empirically (permission denied for Python execution).
- The diff appears to be a proposed change against the committed base, not the working tree — the actual merge outcome may differ from what I analyzed. I based my count analysis on reading the working tree file directly.
- Could not verify whether `context.py:_get_or_create_gate` is the only caller path for gate creation in production use, or whether the `runners.py:_get_or_create_gate` path is used independently.
- Did not have access to the full `_execute_remote_via_gate` method to verify all four PolicyDenied insertion points match exactly.

### Feature Requests
- Include the full working tree state of modified files (not just the committed base) when the diff's base index doesn't match HEAD, to catch merge conflicts.
- Show all callers of modified method signatures to verify integration completeness.
- Include ability to run tests as part of review to verify assertions empirically.

## gemini [CONCERN]

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly updates `AutomationContext` to pass `policy_wire` and `environment` down to the remote runner. Properly handles the new `PolicyDenied` protocol message type in all execution pathways, translating it cleanly to a failed module result.

### src/ftl2/ftl_gate/__main__.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The gate-side enforcement correctly hooks into the request loop, but its fail-open behavior is questionable. If `policy.py` fails to import (e.g., due to a missing third-party dependency like `yaml` on the minimal remote host), `HAS_POLICY` silently becomes `False`. While `SetPolicy` returns an error in this state, the `Hello` message silently ignores the `policy_rules` payload. For a security boundary, receiving policy data but lacking the ability to enforce it should likely be a fatal error or raise a loud warning rather than silently failing open.

### src/ftl2/gate.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly copies `policy.py` and its dependency `exceptions.py` into the gate payload. Properly includes these files in the cache hash calculation to ensure the gate is rebuilt when policy enforcement logic changes.

### src/ftl2/message.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly registers the `PolicyDenied`, `SetPolicy`, and `SetPolicyResult` message types to the protocol definition.

### src/ftl2/policy.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Cleanly introduces serialization and deserialization methods (`to_dict`, `to_wire`, `from_wire`) allowing policies to be transported across the gate protocol.

### src/ftl2/runners.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Properly injects the policy data into the `Hello` handshake and correctly handles `PolicyDenied` gate responses by surfacing a `PolicyDeniedError`.

### tests/test_gate_policy_enforcement.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Extremely thorough test suite. It covers serialization, the gate enforcement check logic, messaging handling, builder bundling, and end-to-end edge cases. Good use of isolated test scenarios.

### tests/test_message.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly updates the total message type count assertion to account for the newly added policy messages.

### Self-Review
**Limitations:** I could not see the full imports at the top of `src/ftl2/policy.py`. If `policy.py` imports a third-party library (like `yaml` for the `from_file` method) at the module level rather than inside the method, the gate will consistently fail to import the policy engine on remote machines where `pyyaml` is not installed, thus silently disabling the enforcement completely due to the `HAS_POLICY` catch block.

PR #97, opened 2026-04-11
Add persistent audit trail for policy decisions

### LINKED ISSUE #71: Add persistent audit trail for policy decisions
## Summary

Policy decisions are logged at debug level only. There is no structured, persistent audit trail of what was denied or permitted, when, by which rule, or with what parameters.

## Problem

When the policy engine evaluates a module invocation:
- **Denied**: `PolicyDeniedError` is raised with the matching rule, but this is only visible if the caller catches and logs it
- **Permitted**: No record at all — `PolicyResult(permitted=True)` is returned and discarded

For the human stakeholder who sets policy, there is no way to answer:
- What did the AI do in the last hour?
- Which policy rules are actually firing?
- Are there attempted violations that indicate a misconfigured AI loop?
- What parameters were passed to permitted operations?

For compliance and enterprise adoption, audit trails are table-stakes.

## Proposal

Add structured audit events for every policy evaluation:

```json
{
  "timestamp": "2026-04-10T13:00:00Z",
  "event": "policy_evaluation",
  "module": "shell",
  "host": "prod-web-01",
  "environment": "prod",
  "params": {"cmd": "restart nginx"},
  "decision": "denied",
  "rule": {"match": {"module": "shell", "environment": "prod"}, "reason": "..."},
  "session_id": "..."
}
```

Options for persistence:
- JSON-lines file (simplest, matches errors-as-data philosophy)
- Structured events on stderr (matches existing event channel)
- Callback hook (most flexible — let the consumer decide storage)

The audit log should capture both denied and permitted decisions. Permitted decisions are more important for forensics — you already know about denials because they raise errors.

## Related

- #70 — Gate-side policy enforcement (gate-side decisions need auditing too)

benthomasson:
Closes #71

### DIFF
--- a/src/ftl2/automation/__init__.py
+++ b/src/ftl2/automation/__init__.py
@@ -77,6 +77,7 @@ async def automation(
     vault_secrets: dict[str, str] | None = None,
     policy: str | None = None,
     environment: str = "",
+    policy_audit: str | None = None,
     ignore_missing_inventory: bool = True,
 ) -> AsyncGenerator[AutomationContext, None]:
     """Create an automation context for running FTL modules.
@@ -157,6 +158,10 @@ async def automation(
                 A matching deny rule raises PolicyDeniedError.
         environment: Environment label for policy matching (e.g., "prod",
                 "staging"). Used by policy rules with environment conditions.
+        policy_audit: Path to JSON-lines file for streaming policy audit events.
+                Each policy evaluation (permitted or denied) is appended as a
+                JSON line immediately after evaluation. Crash-safe — survives
+                process termination. Default is None.
 
     Yields:
         AutomationContext with ftl.module_name() access to all modules
@@ -266,6 +271,7 @@ async def automation(
         vault_secrets=vault_secrets,
         policy=policy,
         environment=environment,
+        policy_audit=policy_audit,
         ignore_missing_inventory=ignore_missing_inventory,
     )
 
--- a/src/ftl2/automation/context.py
+++ b/src/ftl2/automation/context.py
@@ -8,6 +8,7 @@ import asyncio
 import logging
 import os
 import time
+import uuid
 import warnings
 from enum import Enum
 from pathlib import Path
@@ -276,6 +277,7 @@ class AutomationContext:
         vault_secrets: dict[str, str] | None = None,
         policy: str | Path | None = None,
         environment: str = "",
+        policy_audit: str | Path | None = None,
         ignore_missing_inventory: bool = False,
     ):
         """Initialize the automation context.
@@ -357,6 +359,10 @@ class AutomationContext:
             environment: Environment label for policy matching (e.g., "prod",
                 "staging"). Passed to the policy engine for environment-based
                 rules. Default is "" (empty string).
+            policy_audit: Path to a JSON-lines file for streaming policy audit
+                events. Each policy evaluation (permitted or denied) is appended
+                as a JSON line immediately after evaluation. Crash-safe — survives
+                process termination. Default is None (no policy audit file).
         """
         self._enabled_modules = modules
         self._inventory = self._load_inventory(inventory, ignore_missing=ignore_missing_inventory)
@@ -404,6 +410,9 @@ class AutomationContext:
         else:
             self._policy = Policy.empty()
         self._environment = environment
+        self._policy_audit_file = Path(policy_audit) if policy_audit else None
+        self._policy_decisions: list[dict[str, Any]] = []
+        self._session_id = str(uuid.uuid4())
         self._event_handlers: dict[str, dict[str, list]] = {}  # host -> event_type -> [handlers]
         self._proxy = ModuleProxy(self)
         self._hosts_proxy: HostsProxy | None = None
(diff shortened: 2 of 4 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#97
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### `src/ftl2/policy.py:PolicyRule.to_dict`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean serialization method. Returns `decision`, `match`, and `reason` — the three fields of `PolicyRule`. Uses the dataclass fields directly. Tests in `TestPolicyRuleToDict` cover basic, empty match, no-reason, and param-match cases. Used by `_check_policy` to embed the matching rule in audit events.

### `src/ftl2/automation/__init__.py:automation`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Adds `policy_audit` parameter with `None` default. Passes it through to `AutomationContext`. Positioned consistently with other audit-related params. Documentation is accurate. Tests exercise this via `TestPolicyAuditViaDenialPath.test_deny_writes_jsonl_via_automation`.

### `src/ftl2/automation/context.py:__init__` (new fields)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Three new instance variables: `_policy_audit_file` (Path or None), `_policy_decisions` (list), `_session_id` (UUID4 string). Initialization order is correct — placed after `_policy` and `_environment` but before `_event_handlers`. The `Path(policy_audit) if policy_audit else None` pattern correctly handles both `str` and `Path` input types, and `None`. `uuid.uuid4()` generates a unique session ID per context.

### `src/ftl2/automation/context.py:_check_policy`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The method builds an event dict with `"timestamp": time.time()`, then calls `self._emit_event(event)`. However, `_emit_event` **mutates** the passed dict by overwriting `event["timestamp"] = time.time()` (context.py:1110). Since `_policy_decisions` holds a reference to the same dict object, the original timestamp is silently replaced. This is technically consistent (all three destinations end up with the _emit_event timestamp), but it creates a subtle coupling: the timestamp in `_policy_decisions` and the JSONL file depends on whether `_on_event` is registered. When `_on_event is None`, `_emit_event` does nothing and the original `time.time()` survives. When `_on_event` is registered, the timestamp gets overwritten to a slightly later value. This inconsistency is minor (microseconds) but is a code smell that could confuse future maintainers. Also, the method performs file I/O (`_write_policy_audit_line`) on every policy check, including in the happy path — this is by design for crash-safety but worth noting for performance-sensitive workloads.

### `src/ftl2/automation/context.py:_write_policy_audit_line`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly guards on `_policy_audit_file is None`. Uses append mode (`"a"`) for crash-safe writes. Converts epoch timestamp to ISO-8601 using `datetime.fromtimestamp` with explicit UTC timezone. Makes a shallow copy of the event dict before mutating the timestamp field, which properly avoids corrupting the in-memory `_policy_decisions` list's timestamps. The imports are inside the method body (consistent with other methods in this class that do lazy imports). Each line is self-contained valid JSON.

### `src/ftl2/automation/context.py:session_id` (property)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Simple read-only property exposing `_session_id`. Tested in `test_session_id_is_valid_uuid4` and `test_session_id_accessible_on_context`.

### `src/ftl2/automation/context.py:execute` (audit_params change)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Change from `self._check_policy(module_name, params)` to `self._check_policy(module_name, params, audit_params=original_params)`. This ensures injected secrets don't leak into audit events — `original_params` is captured before secret injection. The policy check still evaluates against `params` (post-injection), so rules matching on secret values still work correctly. The audit event records `original_params` (redacted). This is the correct security behavior. Not directly tested in the new test file (would require module execution), but the `test_audit_params_override` test validates the mechanism via `_check_policy` directly.

### `src/ftl2/automation/context.py:_execute_on_host` (audit_params change)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Same pattern as `execute()` — passes `audit_params=original_params` to prevent secret leakage. Correctly placed after secret injection but using the pre-injection params for audit.

### `src/ftl2/automation/context.py:__aexit__` (recording condition)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Changed from `if self._record_file and self._results` to `if self._record_file and (self._results or self._policy_decisions)`. This ensures the recording file is written even when all actions were denied (no execution results but policy decisions exist). This is the correct behavior for compliance auditing — you need to know about denied attempts too. Tested in `test_record_written_with_only_denied_decisions` and `test_deny_writes_record_via_automation`.

### `src/ftl2/automation/context.py:_write_recording` (policy_decisions + session_id)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Adds `policy_decisions` array and `session_id` to the recording JSON. Policy decision timestamps are converted to ISO-8601 (consistent with action timestamps). Makes shallow copies of decision dicts before converting timestamps, preserving the in-memory epoch values. The `session_id` field at the top level enables correlating recording files with JSONL audit files. Tested in `TestPolicyAuditInRecordFile`.

### `tests/test_policy_audit.py`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test file with 426 lines covering: `PolicyRule.to_dict()` (4 tests), direct `_check_policy()` calls (10 tests), JSONL file persistence (4 tests), record file integration (3 tests), and end-to-end denial paths via `automation()` (5 tests). The test strategy is well-designed — it avoids triggering actual module execution (which would require the full Ansible module infrastructure) by testing policy evaluation directly and using denial paths that short-circuit before execution. Tests cover permitted and denied paths, session ID validation, event callbacks, audit_params override for secret safety, environment propagation, host parameter passthrough, multi-event accumulation, JSONL append behavior, ISO-8601 timestamp format, and equivalence group handling. One minor observation: the `test_no_file_written_when_policy_audit_is_none` test checks `list(tmp_path.iterdir()) == []` which is technically testing the absence of side effects in the tmp dir, not specifically the absence of the audit file, but this is fine in practice.

### Self-Review
**Limitations:** - Could not run the test suite to verify tests pass (permission issue with `uv run pytest`). Review is based on code reading only.
- Did not see the full `execute()` and `_execute_on_host()` methods in the modified working tree (only the original + diff), so verified correctness from the diff context lines.
- Did not verify whether any other callers of `_check_policy` exist beyond `execute` and `_execute_on_host` (e.g., in proxy.py or other methods) — if there are others, they would not yet pass `audit_params`.

### Feature Requests
- Include full function bodies of modified functions (not just diff hunks) to verify complete logic without needing to read the original file separately.
- Show all callers of modified methods (like a `grep` for `_check_policy`) to verify integration completeness.
- Auto-run the test suite and include results in the review context.

## gemini [PASS]

### src/ftl2/automation/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Safely adds and propagates the new `policy_audit` parameter down to the `AutomationContext`.

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Well-structured implementation for emitting and persisting policy audit events. The separation of `params` vs `audit_params` correctly ensures that dynamically injected secrets don't end up in the audit log. The audit events successfully track decisions in real-time JSONL formats and also cleanly append to the final session recording.

### src/ftl2/policy.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Adds straightforward and necessary JSON-friendly serialization (`to_dict`) for rules which is seamlessly consumed by the new auditing feature.

### tests/test_policy_audit.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

A very robust and comprehensive test suite. Testing via both direct object manipulation and the `automation` context manager ensures full behavioral coverage without executing underlying (potentially heavy) actual modules.

### Self-Review
**Limitations:** No issue or specification was provided to cross-reference against requirements, so my review relied entirely on internal consistency, structural soundness, and codebase best practices. Additionally, I'm assuming that `_redact_params` handles parameter redaction adequately as the implementation is outside the provided diff scope.

PR #98, opened 2026-04-11
Support runtime policy updates for long-lived permanent gates

### LINKED ISSUE #73: Support runtime policy updates for long-lived permanent gates
## Summary

Policies are loaded once at `AutomationContext` creation via `Policy.from_file()`, `Policy.from_files()`, or `Policy.from_directory()`. There is no mechanism to update policies at runtime. For permanent gates with long-lived contexts, this means policy changes require restarting the automation context.

## Problem

```python
async with automation(policy="rules.yaml") as ftl:
    # Policy is fixed for the lifetime of this context
    # If rules.yaml is updated on disk, the running context doesn't see it
    await ftl.host.shell(cmd="...")  # evaluated against original rules
```

For ephemeral contexts this is fine — next run picks up new rules. For permanent gates (#68) where the context may live for hours or days:
- Policy updates require context restart
- Context restart may require draining in-flight work
- There is a window between policy change and enforcement

## Proposal

Add runtime policy reload:

1. **Manual reload**: `await ftl.reload_policy()` or a `ReloadPolicy` message type
2. **File-watching reload**: Use the existing file watcher capability in the gate to watch policy files and reload on change
3. **For gate-side enforcement (#70)**: An `UpdatePolicy` message that pushes new rules to the gate

The reload should be atomic — the old policy applies until the new one is fully loaded and validated. Invalid policy files should be rejected without affecting the running policy.

## Context

- The `Policy` class already supports loading from files — reload is mechanically simple
- The gate already has a file watcher (`FileChanged` events) that could watch policy files
- Permanent gates make this urgent since they are the primary deployment mode for performance

## Related

- #68 — Gate lifecycle management (long-lived gates need policy updates)
- #70 — Gate-side policy enforcement (gate needs to receive policy updates)

benthomasson:
Closes #73

### DIFF
--- a/src/ftl2/automation/context.py
+++ b/src/ftl2/automation/context.py
@@ -401,8 +401,10 @@ class AutomationContext:
                 data = json.loads(replay_path.read_text())
                 self._replay_actions = data.get("actions", [])
         from ftl2.policy import Policy
+        self._policy_source: Path | None = None
         if policy:
             p = Path(policy)
+            self._policy_source = p
             if p.is_dir():
                 self._policy = Policy.from_directory(p)
             else:
@@ -413,6 +415,7 @@ class AutomationContext:
         self._policy_audit_file = Path(policy_audit) if policy_audit else None
         self._policy_decisions: list[dict[str, Any]] = []
         self._session_id = str(uuid.uuid4())
+        self._policy_watch_task: asyncio.Task | None = None
         self._event_handlers: dict[str, dict[str, list]] = {}  # host -> event_type -> [handlers]
         self._proxy = ModuleProxy(self)
         self._hosts_proxy: HostsProxy | None = None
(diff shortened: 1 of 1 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#98
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [BLOCK]

### `src/ftl2/policy.py:Policy.to_dict`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean serialization. Correctly extracts `decision`, `match`, and `reason` from each `PolicyRule`. The `match` dict is a plain `dict[str, str]` so it serializes without issues. Tests in both `test_policy_reload.py:TestPolicySerialization` and `test_policy_reload_qa.py:TestPolicySerializationQA` verify round-trips including edge cases (param.* keys, empty policies, multiple rules).

### `src/ftl2/policy.py:Policy.from_dict`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The method defaults missing `decision` to `"deny"`, which silently accepts malformed input. While `"deny"` is currently the only valid decision, this means a rule with `{"match": {"module": "shell"}, "reason": "test"}` (no `decision` key) will be accepted silently. This is fine for internal use (protocol transmission where you control the format), but could mask bugs if used for user-authored data. Tests cover validation of invalid decisions and empty/None rules lists. Minor concern, not a blocker since `PolicyRule.__post_init__` validates the decision value.

### `src/ftl2/automation/context.py:AutomationContext.reload_policy`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correct load-then-swap pattern: the new policy is fully loaded and validated before `self._policy` is reassigned. If loading raises (invalid YAML, bad decision values, missing file), the old policy remains unchanged — verified by `test_reload_atomic_on_invalid_file`. Updates `_policy_source` only when an explicit `path` is provided, so subsequent pathless reloads use the new source. All error paths tested: no source, missing file, invalid content.

### `src/ftl2/automation/context.py:AutomationContext.watch_policy`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The implementation is functional and well-tested (change detection, error survival, idempotency, `__aexit__` cleanup). Two minor concerns:

1. **Continuous retry on invalid files**: When reload fails, `self._policy_mtimes` is NOT updated, so the watcher retries every `interval` seconds with `exc_info=True` logging. This is noisy but arguably correct (it will succeed once the file is fixed). The tests confirm this behavior (`test_watch_survives_invalid_file`).

2. **`_policy_mtimes` defined outside `__init__`**: The instance attribute is created inside `watch_policy()` rather than initialized in `__init__`. This is a minor style concern — it's only accessed by `_watch_loop` and `_snapshot_mtimes`, so it won't cause AttributeError in practice.

### `src/ftl2/automation/context.py:AutomationContext._snapshot_mtimes`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Mirrors the file discovery logic in `Policy.from_directory()` (same suffix filter, same `sorted()` + `is_file()` pattern). Handles `FileNotFoundError` for single files gracefully. For directories, if a file is deleted between `iterdir()` and `stat()`, a TOCTOU race could raise `FileNotFoundError` — but this is inherent to polling and would just be caught by the watcher's `except Exception` handler. No direct unit tests for this private method, but exercised through the watch tests.

### `src/ftl2/automation/context.py:AutomationContext.__aexit__` (cleanup)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

`unwatch_policy()` is called at the top of `__aexit__`, before other cleanup. `task.cancel()` is fire-and-forget; the `CancelledError` propagates through `asyncio.sleep()` in `_watch_loop` and asyncio suppresses it since the task was explicitly cancelled. Test `test_aexit_cancels_watcher` confirms the task is set to None after exit.

### `src/ftl2/message.py:GateProtocol.MESSAGE_TYPES`
**Verdict:** BLOCK
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

Adding `UpdatePolicy` and `UpdatePolicyResult` to the `MESSAGE_TYPES` set is correct as protocol reservation for future work (#70). However, `tests/test_message.py:19` asserts `len(protocol.MESSAGE_TYPES) == 25` — the pre-change count. With 2 new entries the count is now 27, so **this test will fail**. The test file is not updated in the diff. Additionally, these message types are declared but have no handler implementation on either side (controller or gate) — the comment notes "future: #70", which is fine for reservation but the missing test update is a blocker.

### Self-Review
**Limitations:** - Could not verify the full `__init__` signature to confirm `policy` parameter type and whether `_policy_source` could interact with any other path that sets `self._policy`
- Did not verify whether any existing callers of `AutomationContext` would be affected by the new `policy` property (name collision with the init parameter)
- Could not run the test suite to confirm the MESSAGE_TYPES count breakage — inference based on counting set members

### Feature Requests
- Show full `__init__` and `__aexit__` methods for modified classes, not just diff hunks
- Auto-detect test files that assert on counts/sizes of modified collections and flag potential breakage
- Flag when a test file appears in git status modifications but is absent from the review diff

## gemini [BLOCK]

**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

The `_watch_loop` function in `watch_policy` contains a bug that will cause severe log spamming. If `reload_policy()` raises an exception (e.g., due to an invalid YAML file), `self._policy_mtimes` is never updated. On the next iteration (`interval` seconds later), `current != self._policy_mtimes` will still be true, causing the loop to infinitely try to reload the broken file and log exceptions every second until the file is modified again. `self._policy_mtimes` must be updated regardless of whether the reload succeeds or fails. Additionally, `_policy_mtimes` is defined outside of `__init__`.

### src/ftl2/message.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

Safely adds `UpdatePolicy` and `UpdatePolicyResult` to `GATE_COMMANDS`. The integration is marked partial as this seems to be laying groundwork for a future feature (Issue #70) but nothing actively uses it yet.

### src/ftl2/policy.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

The `from_dict` docstring explicitly claims it "Validates all rules before creating the Policy. If any rule is invalid, raises ValueError". However, the implementation simply passes `entry.get("match", {})` and other dict values directly to `PolicyRule()`. Unless `PolicyRule` performs strict type-checking and validation (and raises `ValueError`), this docstring is inaccurate and the method could silently accept malformed rules.

### Self-Review
**Limitations:** Could not view the `PolicyRule` class implementation to verify if it enforces the validation contract claimed by `Policy.from_dict()`. No test files were included in the diff to verify coverage.

### Feature Requests
- Include the implementation of classes/functions referenced in the changes (e.g., `PolicyRule`) to verify implicit behavior and validation logic.
- Ensure test files modified or added as part of the commit are included in the diff snippet for accurate test coverage evaluation.

PR #113, opened 2026-04-18
Support host range patterns in inventory files

### LINKED ISSUE #106: Support host range patterns in inventory files
## Problem

FTL2 does not support host range patterns in inventory files.

## Expected Behavior

Support Ansible host range syntax in both YAML and INI inventory formats:

### Numeric ranges
```yaml
webservers:
  hosts:
    www[01:50].example.com:      # www01 through www50 (inclusive)
```

### Numeric ranges with stride
```yaml
webservers:
  hosts:
    www[01:50:2].example.com:    # www01, www03, www05, ..., www49
```

### Alphabetic ranges
```yaml
databases:
  hosts:
    db-[a:f].example.com:        # db-a through db-f
```

### Leading zeros
Numeric ranges preserve leading zeros from the start value. `www[01:50]` produces `www01` through `www50`, not `www1` through `www50`.

## Why This Matters

Host ranges are commonly used in production inventories that manage many similar hosts (web server farms, database clusters, etc.). The porting flywheel needs to handle these without manual expansion.

benthomasson:
## Summary

- Adds `expand_host_range()` for Ansible-style patterns: `www[01:50].example.com`
- Numeric ranges with leading zero preservation, alphabetic ranges, stride
- Cartesian product for multiple bracket groups: `rack[1:2]-node[a:b]`
- Expansion at YAML parse time; host vars propagate to all expanded hosts
- 11 new tests (unit + integration through YAML loading)

## Test plan

- [x] All 47 inventory tests pass
- [x] Full suite: 1816 passed, 0 failed, 8 skipped
- [x] CI validates on Ubuntu runner

Closes #106

### DIFF
--- a/src/ftl2/inventory.py
+++ b/src/ftl2/inventory.py
@@ -247,8 +247,6 @@ def _host_from_vars(host_name: str, host_data: dict[str, Any]) -> HostConfig:
     )
 
 
-_NUMERIC_RANGE_RE = re.compile(r"\[([0-9]+):([0-9]+)(?::([0-9]+))?\]")
-_ALPHA_RANGE_RE = re.compile(r"\[([a-zA-Z]):([a-zA-Z])(?::([0-9]+))?\]")
 _RANGE_RE = re.compile(
     r"\[([0-9]+):([0-9]+)(?::([0-9]+))?\]"
     r"|"
--- a/src/ftl2/inventory.py
+++ b/src/ftl2/inventory.py
@@ -247,11 +247,23 @@ def _host_from_vars(host_name: str, host_data: dict[str, Any]) -> HostConfig:
     )
 
 
-_RANGE_RE = re.compile(r"\[([a-zA-Z0-9]+:[a-zA-Z0-9]+(?::[0-9]+)?)\]")
+_NUMERIC_RANGE_RE = re.compile(r"\[([0-9]+):([0-9]+)(?::([0-9]+))?\]")
+_ALPHA_RANGE_RE = re.compile(r"\[([a-zA-Z]):([a-zA-Z])(?::([0-9]+))?\]")
+_RANGE_RE = re.compile(
+    r"\[([0-9]+):([0-9]+)(?::([0-9]+))?\]"
+    r"|"
+    r"\[([a-zA-Z]):([a-zA-Z])(?::([0-9]+))?\]"
+)
 
 
 def expand_host_range(pattern: str) -> list[str]:
-    """Expand Ansible host range patterns like ``www[01:50].example.com``."""
+    """Expand Ansible host range patterns like ``www[01:50].example.com``.
+
+    Supports numeric ranges (``[01:50]``), alphabetic ranges (``[a:f]``),
+    and stride (``[01:50:2]``).  Multiple bracket groups produce a cartesian
+    product.  Brackets that don't match valid range syntax are left as
+    literal characters.
+    """
     matches = list(_RANGE_RE.finditer(pattern))
     if not matches:
         return [pattern]
@@ -260,15 +272,23 @@ def expand_host_range(pattern: str) -> list[str]:
     last_end = 0
     for m in matches:
         segments.append([pattern[last_end : m.start()]])
-        parts = m.group(1).split(":")
-        start, end = parts[0], parts[1]
-        stride = int(parts[2]) if len(parts) > 2 else 1
 
-        if start.isdigit() and end.isdigit():
-            width = len(start)
+        if m.group(1) is not None:
+            # Numeric range
+            start, end = m.group(1), m.group(2)
+            stride = int(m.group(3)) if m.group(3) else 1
+            if stride == 0:
+                raise ValueError(f"Invalid host range stride of 0 in '{pattern}'")
+            width = max(len(start), len(end))
             vals = [str(i).zfill(width) for i in range(int(start), int(end) + 1, stride)]
         else:
+            # Alpha range
+            start, end = m.group(4), m.group(5)
+            stride = int(m.group(6)) if m.group(6) else 1
+            if stride == 0:
+                raise ValueError(f"Invalid host range stride of 0 in '{pattern}'")
             vals = [chr(c) for c in range(ord(start), ord(end) + 1, stride)]
+
         segments.append(vals)
         last_end = m.end()
 
--- a/src/ftl2/inventory.py
+++ b/src/ftl2/inventory.py
@@ -247,7 +247,7 @@ def _host_from_vars(host_name: str, host_data: dict[str, Any]) -> HostConfig:
     )
 
 
-_RANGE_RE = re.compile(r"\[([^\]]+)\]")
+_RANGE_RE = re.compile(r"\[([a-zA-Z0-9]+:[a-zA-Z0-9]+(?::[0-9]+)?)\]")
 
 
 def expand_host_range(pattern: str) -> list[str]:
--- a/src/ftl2/inventory.py
+++ b/src/ftl2/inventory.py
@@ -4,8 +4,10 @@ This module provides typed inventory management with dataclasses, replacing
 dictionary-based inventory structures with strongly-typed classes.
 """
 
+import itertools
 import json
 import os
+import re
 import subprocess
 import sys
 from dataclasses import dataclass, field
@@ -166,7 +168,8 @@ def _process_group(
             for host_name, host_data in group_data["hosts"].items():
                 if not isinstance(host_data, dict):
                     host_data = {}
-                group.add_host(_host_from_vars(host_name, host_data))
+                for expanded in expand_host_range(host_name):
+                    group.add_host(_host_from_vars(expanded, host_data))
 
         if "vars" in group_data and isinstance(group_data["vars"], dict):
             group.vars.update(group_data["vars"])
@@ -242,6 +245,35 @@ def _host_from_vars(host_name: str, host_data: dict[str, Any]) -> HostConfig:
         ),
         vars={k: v for k, v in host_data.items() if k not in standard_fields},
     )
+
+
+_RANGE_RE = re.compile(r"\[([^\]]+)\]")
+
+
+def expand_host_range(pattern: str) -> list[str]:
+    """Expand Ansible host range patterns like ``www[01:50].example.com``."""
+    matches = list(_RANGE_RE.finditer(pattern))
+    if not matches:
+        return [pattern]
+
+    segments: list[list[str]] = []
+    last_end = 0
+    for m in matches:
+        segments.append([pattern[last_end : m.start()]])
+        parts = m.group(1).split(":")
+        start, end = parts[0], parts[1]
+        stride = int(parts[2]) if len(parts) > 2 else 1
+
+        if start.isdigit() and end.isdigit():
+            width = len(start)
+            vals = [str(i).zfill(width) for i in range(int(start), int(end) + 1, stride)]
+        else:
+            vals = [chr(c) for c in range(ord(start), ord(end) + 1, stride)]
+        segments.append(vals)
+        last_end = m.end()
+
+    segments.append([pattern[last_end:]])
+    return ["".join(combo) for combo in itertools.product(*segments)]
 
 
 def load_inventory_json(data: dict[str, Any], require_hosts: bool = True) -> Inventory:
--- a/tests/test_inventory.py
+++ b/tests/test_inventory.py
@@ -5,6 +5,8 @@ import tempfile
 from pathlib import Path
 from unittest.mock import patch
 
+import pytest
+
 from ftl2.inventory import (
     HostGroup,
     Inventory,
@@ -782,9 +784,27 @@ class TestExpandHostRange:
         """Brackets without colon are not range syntax — pass through literally."""
         assert expand_host_range("host[tag]") == ["host[tag]"]
 
+    def test_multi_char_alpha_passthrough(self):
+        """Multi-character alpha brackets are not valid ranges — pass through."""
+        assert expand_host_range("host[tag:value]") == ["host[tag:value]"]
+
+    def test_ipv6_like_passthrough(self):
+        """IPv6-like brackets are not valid ranges — pass through."""
+        assert expand_host_range("host[2001:db8]") == ["host[2001:db8]"]
+
     def test_descending_range_empty(self):
         """Descending range produces no values — returns empty list."""
         assert expand_host_range("host[5:1]") == []
+
+    def test_zero_stride_raises(self):
+        """Stride of 0 raises ValueError."""
+        with pytest.raises(ValueError, match="stride of 0"):
+            expand_host_range("host[1:5:0]")
+
+    def test_width_from_longer_side(self):
+        """Zero-padding uses the wider of start/end."""
+        result = expand_host_range("host[1:003]")
+        assert result == ["host001", "host002", "host003"]
 
 
 class TestHostRangeIntegration:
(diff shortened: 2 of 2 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#113
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/inventory.py:expand_host_range
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** PARTIAL

The function is correct for well-formed inputs but has two issues:

1. **No input validation on malformed patterns.** A bracket expression without a colon (e.g. `host[foo]`) will crash with `IndexError` at line 264 (`parts[1]`). The regex `_RANGE_RE` matches any `[...]` content, not just range syntax. A host like `host[tag]` in inventory would blow up rather than being treated as a literal or raising a clear error.

2. **Only wired into YAML inventory path.** `load_inventory_json` (line 333) and `load_inventory_script` do not call `expand_host_range`. While JSON/script inventories from `ansible-inventory --list` typically return already-expanded host lists, this is an inconsistency worth noting. Ansible's own JSON format doesn't use range patterns, so this is more of a design observation than a bug.

3. **Zero-width edge case with `isdigit()`.** `"0".isdigit()` is `True`, so `[0:0]` works, but `[00:00]` will produce `["00"]` — correct behavior. No bug here.

### src/ftl2/inventory.py:_process_group (integration point)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The integration in `_process_group` is clean. Each expanded hostname gets the same `host_data` dict, which matches Ansible's behavior — range-expanded hosts share the vars declared on the pattern line. The loop replaces the single `add_host` call correctly.

### tests/test_inventory.py:TestExpandHostRange
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good coverage of numeric ranges, alpha ranges, zero-padding, strides, cartesian product of multiple ranges, single-value ranges, and passthrough. Missing: a test for malformed bracket expressions (e.g., `host[foo]` without a colon) — this would expose the `IndexError` bug noted above.

### tests/test_inventory.py:TestHostRangeIntegration
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Integration tests verify the end-to-end path through YAML inventory loading for numeric ranges, alpha ranges, and mixed range+plain hosts. Correctly asserts that `host_data` (like `ansible_user`) is propagated to all expanded hosts.

### Self-Review
**Limitations:** Could not verify whether Ansible itself supports bracket expressions without colons in hostnames (which would make this a real-world concern vs. theoretical). Could not run tests to confirm they pass. Did not see the full INI inventory path — only YAML was visible, so cannot confirm whether INI parsing also needs this.

### Feature Requests
- Include Ansible's own documentation or behavior reference for host range patterns to verify edge case handling against the upstream spec
- Show callers of modified/new functions across the codebase to verify full integration coverage

## gemini [CONCERN]

### src/ftl2/inventory.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The `expand_host_range` implementation is fragile and crashes or behaves unexpectedly on invalid or non-range bracket syntax:
1. Unhandled `IndexError`: A string with brackets but no colons (e.g., `"host[1]"`) causes `parts[1]` to throw an `IndexError`.
2. Unhandled `TypeError`: Multi-character alpha ranges (e.g., `"host[aa:bb]"`) cause `ord("aa")` to throw a `TypeError`.
3. Silent Drops: Descending ranges (e.g., `"host[3:1]"`) or mismatched case (e.g., `"host[a:Z]"`) result in an empty `range()`, causing `itertools.product` to yield 0 items, silently dropping the host from the inventory rather than raising a clear error.
Consider tightening `_RANGE_RE` (e.g., `r"\[([a-zA-Z0-9]+:[a-zA-Z0-9]+(?::[0-9]+)?)\]"`) so it only matches valid ranges, and adding explicit validation/error messages for the rest.

### tests/test_inventory.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good test coverage for the happy paths, including numeric/alpha ranges, strides, zero-padding, multiple ranges, and YAML inventory integration. However, it lacks test cases for malformed/edge-case patterns (like brackets without colons or descending ranges) which currently trigger unhandled exceptions in the source code.

### Self-Review
**Limitations:** None. Context was sufficient to evaluate the standalone range expansion logic and its integration into the YAML inventory parser.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#113
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### `src/ftl2/inventory.py:expand_host_range`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** PARTIAL

**Edge case bugs:**

1. **Mixed alpha-numeric ranges** — The regex `[a-zA-Z0-9]+:[a-zA-Z0-9]+` matches `[a:3]` or `[1:b]`. The code branches on `start.isdigit() and end.isdigit()`, so if only one is a digit (`[a:3]`), it falls through to `chr(c) for c in range(ord('a'), ord('3') + 1)` which silently produces garbage characters (ASCII 97 through 51 — an empty range since 97 > 51, but reversed cases like `[3:a]` would emit non-printable characters). The regex should enforce that both sides are the same type (all digits or all single alpha chars).

2. **Multi-character alpha values** — The regex matches `[ab:cd]`. Since both `ab` and `cd` are non-digit, the code calls `ord('ab')` which raises `TypeError: ord() expected a character, but string of length 2 found`. This is a **runtime crash** on valid-looking input. Ansible restricts alpha ranges to single characters; the regex should too: `([a-zA-Z]:[a-zA-Z](?::[0-9]+)?)` for alpha or `([0-9]+:[0-9]+(?::[0-9]+)?)` for numeric.

3. **Zero stride** — `host[1:5:0]` matches the regex and would cause `range(1, 6, 0)` → `ValueError: range() arg 3 must not be zero`. No guard against stride=0.

4. **Descending range returns empty list** — `expand_host_range("host[5:1]")` returns `[]` (empty list). This means a host entry with a descending range silently disappears from inventory. This is tested and documented, but Ansible actually raises an error for this case. Silently dropping hosts is dangerous — a typo like `[10:01]` (forgot leading zero) would produce zero hosts with no warning.

5. **`width` uses `len(start)` only** — For `[1:003]`, width would be 1 (from `start="1"`), producing `["1", "2", "3"]` instead of `["001", "002", "003"]`. Ansible uses `max(len(start), len(end))`. Minor but worth noting.

### `src/ftl2/inventory.py:_RANGE_RE`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The regex `r"\[([a-zA-Z0-9]+:[a-zA-Z0-9]+(?::[0-9]+)?)\]"` is overly permissive. It matches multi-character alpha ranges (crash), mixed alpha/numeric ranges (garbage), and doesn't distinguish between numeric and alpha patterns at the regex level. Ansible's own implementation is more restrictive. The commit message ("Tighten host range regex to only match valid range syntax") suggests this was already iterated on, but it still has gaps.

### `src/ftl2/inventory.py:_process_group` (integration point)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** PARTIAL

The integration into `_process_group` is clean — each host key is expanded via `expand_host_range` and all expanded hosts share the same `host_data`. This is correct behavior. However, `load_inventory_json` does NOT call `expand_host_range` — JSON inventory host names are not expanded. This is arguably correct (JSON format comes from `ansible-inventory --list` which pre-expands ranges), but it's worth noting the asymmetry.

### `tests/test_inventory.py:TestExpandHostRange`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Good coverage of the happy path: numeric, alpha, stride, cartesian product, passthrough, single-value. The `test_descending_range_empty` documents the empty-list behavior. Missing tests for the crash cases identified above: multi-char alpha, mixed alpha/numeric, zero stride.

### `tests/test_inventory.py:TestHostRangeIntegration`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Integration tests verify the full path from YAML file → `load_inventory` → expanded hosts. Tests cover numeric ranges, alpha ranges, and mixed range + plain hosts. Good coverage of the integration seam.

### Self-Review
**Limitations:** Could not execute Python to verify the edge case behavior empirically — analysis is from code reading. Could not see the full git history to check if the tightened regex (commit cdd2db3) addressed some of these already. Did not check if Ansible's own implementation has similar behavior for descending ranges.

### Feature Requests
- Include Ansible's own `expand_host_range` behavior as reference when reviewing code that emulates Ansible functionality
- Allow Python execution in review context to verify edge case claims

## gemini [BLOCK]

### src/ftl2/inventory.py:expand_host_range
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The implementation crashes with a `TypeError` on inputs like `host[tag:value]` or IPv6-like segments `host[2001:db8]`. The regex `[a-zA-Z0-9]+:[a-zA-Z0-9]+` matches these multi-character strings. Because they are not entirely digits, the logic falls through to the `else` block and calls `ord(start)`. Since `ord()` only accepts single-character strings, providing a string of length > 1 raises an unhandled exception, breaking the inventory parser. Furthermore, descending ranges (e.g., `host[5:1]`) return an empty list, silently dropping the host from the inventory entirely rather than processing a descending sequence or raising a validation error.

### tests/test_inventory.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The tests cover the happy paths for range expansion but fail to include negative or edge cases such as multi-character alphabetic bounds (e.g., `[aa:bb]`) or non-range bracketed values containing colons (e.g., `[tag:value]`), which would have exposed the `TypeError`. Additionally, `test_descending_range_empty` explicitly enforces that descending ranges yield an empty list—which causes silent host dropping rather than producing a descending list (like Ansible does) or raising an error for a typo.

### Self-Review
**Limitations:** None — the provided diff and full function bodies were sufficient to identify the crash scenario and analyze the regex behavior.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#113
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/inventory.py:expand_host_range
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** PARTIAL

The function is well-implemented with one notable issue: **descending ranges return an empty list** (`host[5:1]` → `[]`), and when this happens through the inventory loader at line 171, the host entry silently vanishes. A typo like `www[50:01]` would produce an empty inventory with no warning. Ansible itself produces an empty range too, so this is arguably compatible, but it's a footgun worth flagging. Additionally, `_NUMERIC_RANGE_RE` and `_ALPHA_RANGE_RE` (lines 250-251) are compiled but never used — only `_RANGE_RE` is used. These are dead code.

### src/ftl2/inventory.py:_process_group (integration)
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** PARTIAL

Host range expansion is wired into `_process_group` (YAML path) but **not** into `load_inventory_json` (line 353). JSON inventory hosts could also contain range patterns — Ansible supports ranges in both formats. If this is intentional (JSON comes from `--list` which already has expanded hosts), it should be documented. The `host_data` dict is shared across all expanded hosts (line 172), which is correct since all hosts from a range pattern should inherit the same vars. However, this means they share the same dict object — currently safe because `_host_from_vars` creates a new dict for `vars`.

### src/ftl2/inventory.py (dead code)
VERDICT: CONCERN
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: N/A
REASONING: `_NUMERIC_RANGE_RE` (line 250) and `_ALPHA_RANGE_RE` (line 251) are compiled regex patterns that are never referenced anywhere in the codebase. They appear to be intermediate development artifacts superseded by the combined `_RANGE_RE`. They should be removed.

---

### tests/test_inventory.py:TestExpandHostRange
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive unit tests covering numeric ranges, alphabetic ranges, stride, zero-padding, cartesian product, passthrough for non-range brackets, descending range, zero-stride error, and width inference. Edge cases are well-chosen — the `test_ipv6_like_passthrough` and `test_multi_char_alpha_passthrough` tests verify that the regex doesn't over-match. One minor gap: no test for mixed numeric + alpha ranges in a single pattern (e.g., `rack[1:2]-shelf[a:b]`).

### tests/test_inventory.py:TestHostRangeIntegration
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Integration tests verify the end-to-end path from YAML file → `load_inventory` → expanded hosts. Tests cover numeric ranges with host vars propagation, alphabetic ranges, and mixed range + plain hosts. Good coverage of the integration point.

### Self-Review
**Limitations:** Could not verify whether Ansible's JSON inventory format ever contains unexpanded range patterns (would need to check Ansible source). Could not verify whether there are other callers that construct `_process_group` input besides the YAML loader.

### Feature Requests
- Show all callers of modified functions to verify complete integration
- Flag compiled regex patterns that are never used (dead code detection)

## gemini [CONCERN]

### src/ftl2/inventory.py:expand_host_range
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The function intentionally returns an empty list for descending ranges (e.g., `[5:1]`), whereas standard Ansible supports descending range expansion. This means hosts defined with a descending range will be silently dropped. Additionally, the module-level regex variables `_NUMERIC_RANGE_RE` and `_ALPHA_RANGE_RE` are defined but never used in the function since it relies entirely on `_RANGE_RE`.

### src/ftl2/inventory.py:_process_group
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly integrates `expand_host_range` into the inventory parsing flow. Because `_host_from_vars` extracts values and creates a new dictionary for variables, reusing `host_data` across multiple generated hosts is perfectly safe and will not lead to shared state mutation.

### tests/test_inventory.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Adds a thorough set of unit tests covering single and multi-range cartesian expansions, numeric and alpha strides, correct zero-padding based on width, and invalid syntax fallbacks. The YAML integration tests ensure the feature is fully wired into the inventory loader.

### Self-Review
**Limitations:** None. The diff is self-contained and the scope is clear.

PR #115, opened 2026-04-18
Load host_vars/ and group_vars/ directories alongside inventory

### LINKED ISSUE #108: Load host_vars/ and group_vars/ directories alongside inventory
## Problem

FTL2 does not load `host_vars/` and `group_vars/` directories alongside inventory files.

## Expected Behavior

When loading an inventory file, also scan for and load variable files from adjacent directories:

```
inventory/
  hosts.yml
  group_vars/
    all.yml                    # variables for all hosts
    webservers.yml             # variables for webservers group
    raleigh/                   # directory form
      db_settings.yml
      cluster_settings.yml
  host_vars/
    foosball.yml               # variables for host foosball
```

### Rules

- Valid file extensions: `.yml`, `.yaml`, `.json`, or no extension
- Files in directories are loaded in lexicographical order
- Variables from these files merge with inline inventory variables
- `group_vars/` and `host_vars/` are relative to the inventory source
- Variable precedence follows standard rules (all < parent group < child group < host)

## Why This Matters

Separating variables into `host_vars/` and `group_vars/` is the recommended Ansible practice for organizing complex inventories. Production deployments commonly use this pattern, especially for sensitive variables managed with Ansible Vault.

benthomasson:
Here's the PR description:

## Summary

Add support for loading `host_vars/` and `group_vars/` directories alongside inventory files, matching standard Ansible inventory layout conventions. Variables from external files are merged with inline inventory variables, with support for YAML, JSON, and no-extension file formats. Closes #108

## Changes

- Added `_load_vars_file()` and `_load_vars_dir()` helpers to load and recursively merge variable files (`.yml`, `.yaml`, `.json`, or no extension) in lexicographical order
- Added `_apply_external_vars()` to scan for `group_vars/` and `host_vars/` directories relative to the inventory file and apply variables to matching groups and hosts
- Integrated external vars loading into all inventory format paths (YAML, JSON, INI, and executable script)
- Standard Ansible connection fields (`ansible_host`, `ansible_user`, etc.) from `host_vars/` are promoted to host attributes rather than stored in the generic `vars` dict
- Files with invalid extensions (e.g., `.txt`) are silently ignored
- Added `group_vars/all` support, creating the `all` group if it doesn't already exist

## Test Plan

- [x] `group_vars/` loading from `.yml`, `.json`, and no-extension files
- [x] `group_vars/` directory form with lexicographical merge ordering
- [x] `group_vars/all` applies to the `all` group
- [x] `host_vars/` loading with standard field promotion (`ansible_user`, etc.)
- [x] Invalid file extensions are ignored
- [x] External vars override inline inventory vars
- [x] No `group_vars/` or `host_vars/` directories present (no-op)
- [x] Works with INI inventory format
- [x] Run `pytest tests/test_inventory.py::TestExternalVars`

### DIFF
--- a/src/ftl2/inventory.py
+++ b/src/ftl2/inventory.py
@@ -99,6 +99,61 @@ class Inventory:
         return result
 
 
+_VALID_VARS_EXTENSIONS = {".yml", ".yaml", ".json", ""}
+
+
+def _load_vars_file(path: Path) -> dict[str, Any]:
+    content = path.read_text()
+    if path.suffix == ".json":
+        return json.loads(content) or {}
+    return yaml.safe_load(content) or {}
+
+
+def _load_vars_dir(dirpath: Path) -> dict[str, Any]:
+    if dirpath.is_file():
+        if dirpath.suffix not in _VALID_VARS_EXTENSIONS:
+            return {}
+        return _load_vars_file(dirpath)
+    result: dict[str, Any] = {}
+    for entry in sorted(dirpath.iterdir()):
+        result.update(_load_vars_dir(entry))
+    return result
+
+
+def _apply_external_vars(inventory: Inventory, inventory_path: Path) -> None:
+    base = inventory_path.parent
+    standard_fields = {
+        "ansible_host", "ansible_port", "ansible_user",
+        "ansible_connection", "ansible_python_interpreter",
+    }
+
+    group_vars_dir = base / "group_vars"
+    if group_vars_dir.is_dir():
+        for entry in sorted(group_vars_dir.iterdir()):
+            group_name = entry.stem
+            group = inventory.get_group(group_name)
+            if group is None and group_name == "all":
+                group = HostGroup(name="all")
+                inventory.add_group(group)
+            if group is None:
+                continue
+            group.vars.update(_load_vars_dir(entry))
+
+    host_vars_dir = base / "host_vars"
+    if host_vars_dir.is_dir():
+        all_hosts = inventory.get_all_hosts()
+        for entry in sorted(host_vars_dir.iterdir()):
+            host_name = entry.stem
+            host = all_hosts.get(host_name)
+            if host is None:
+                continue
+            loaded = _load_vars_dir(entry)
+            for field in standard_fields:
+                if field in loaded:
+                    setattr(host, field, loaded.pop(field))
+            host.vars.update(loaded)
+
+
 def load_inventory(inventory_file: str | Path, require_hosts: bool = True) -> Inventory:
     """Load inventory from a file, auto-detecting the format.
 
@@ -132,7 +187,9 @@ def load_inventory(inventory_file: str | Path, require_hosts: bool = True) -> In
 
     # Executable script — run with --list
     if os.access(path, os.X_OK) and path.suffix not in (".yml", ".yaml", ".json", ".ini", ".cfg"):
-        return load_inventory_script(path, require_hosts=require_hosts)
+        inv = load_inventory_script(path, require_hosts=require_hosts)
+        _apply_external_vars(inv, path)
+        return inv
 
     content = path.read_text()
 
@@ -140,17 +197,23 @@ def load_inventory(inventory_file: str | Path, require_hosts: bool = True) -> In
     stripped = content.lstrip()
     if stripped.startswith("{"):
         data = json.loads(content)
-        return load_inventory_json(data, require_hosts=require_hosts)
+        inv = load_inventory_json(data, require_hosts=require_hosts)
+        _apply_external_vars(inv, path)
+        return inv
 
     # INI — detect by extension or content heuristic (skip YAML extensions)
     if path.suffix in (".ini", ".cfg") or (
         path.suffix not in (".yml", ".yaml") and _is_ini_content(stripped)
     ):
-        return load_inventory_ini(content, require_hosts=require_hosts)
+        inv = load_inventory_ini(content, require_hosts=require_hosts)
+        _apply_external_vars(inv, path)
+        return inv
 
     # YAML — existing format
     data = yaml.safe_load(content) or {}
-    return _load_inventory_yaml(data, require_hosts=require_hosts)
+    inv = _load_inventory_yaml(data, require_hosts=require_hosts)
+    _apply_external_vars(inv, path)
+    return inv
 
 
 def _process_group(
--- a/src/ftl2/inventory.py
+++ b/src/ftl2/inventory.py
@@ -110,17 +110,37 @@ def _load_vars_file(path: Path) -> dict[str, Any]:
 
 
 def _load_vars_dir(dirpath: Path) -> dict[str, Any]:
+    """Load vars from a file or directory.
+
+    If *dirpath* is a file, load it directly (any extension — the caller
+    decides whether this entry is valid).  If it is a directory, iterate
+    entries in sorted order, filtering by ``_VALID_VARS_EXTENSIONS`` to
+    skip stray files like ``README.txt``.
+    """
     if dirpath.is_file():
-        if dirpath.suffix not in _VALID_VARS_EXTENSIONS:
-            return {}
         return _load_vars_file(dirpath)
     result: dict[str, Any] = {}
     for entry in sorted(dirpath.iterdir()):
+        if entry.is_file() and entry.suffix not in _VALID_VARS_EXTENSIONS:
+            continue
         result.update(_load_vars_dir(entry))
     return result
 
 
+def _vars_entry_name(entry: Path) -> str:
+    """Extract the group/host name from a vars directory entry.
+
+    Only strips suffixes that are valid vars extensions (.yml, .yaml, .json).
+    This preserves FQDN hostnames like ``db.example.com`` when stored as
+    extensionless files or directories.
+    """
+    if entry.suffix in (".yml", ".yaml", ".json"):
+        return entry.stem
+    return entry.name
+
+
 def _apply_external_vars(inventory: Inventory, inventory_path: Path) -> None:
+    """Load group_vars/ and host_vars/ directories adjacent to the inventory file."""
     base = inventory_path.parent
     standard_fields = {
         "ansible_host", "ansible_port", "ansible_user",
@@ -130,7 +150,7 @@ def _apply_external_vars(inventory: Inventory, inventory_path: Path) -> None:
     group_vars_dir = base / "group_vars"
     if group_vars_dir.is_dir():
         for entry in sorted(group_vars_dir.iterdir()):
-            group_name = entry.stem
+            group_name = _vars_entry_name(entry)
             group = inventory.get_group(group_name)
             if group is None and group_name == "all":
                 group = HostGroup(name="all")
@@ -143,14 +163,14 @@ def _apply_external_vars(inventory: Inventory, inventory_path: Path) -> None:
     if host_vars_dir.is_dir():
         all_hosts = inventory.get_all_hosts()
         for entry in sorted(host_vars_dir.iterdir()):
-            host_name = entry.stem
+            host_name = _vars_entry_name(entry)
             host = all_hosts.get(host_name)
             if host is None:
                 continue
             loaded = _load_vars_dir(entry)
-            for field in standard_fields:
-                if field in loaded:
-                    setattr(host, field, loaded.pop(field))
+            for field_name in standard_fields:
+                if field_name in loaded:
+                    setattr(host, field_name, loaded.pop(field_name))
             host.vars.update(loaded)
 
 
--- a/src/ftl2/inventory.py
+++ b/src/ftl2/inventory.py
@@ -410,16 +410,6 @@ def _host_from_vars(host_name: str, host_data: dict[str, Any]) -> HostConfig:
     Returns:
         HostConfig with standard fields extracted and remainder in vars
     """
-    _STANDARD_HOST_FIELDS = {
-        "ansible_host",
-        "ansible_port",
-        "ansible_user",
-        "ansible_connection",
-        "ansible_python_interpreter",
-        "ansible_become",
-        "ansible_become_user",
-    }
-
     return HostConfig(
         name=host_name,
         ansible_host=host_data.get("ansible_host", host_name),
--- a/src/ftl2/inventory.py
+++ b/src/ftl2/inventory.py
@@ -145,6 +145,7 @@ def _apply_external_vars(inventory: Inventory, inventory_path: Path) -> None:
     standard_fields = {
         "ansible_host", "ansible_port", "ansible_user",
         "ansible_connection", "ansible_python_interpreter",
+        "ansible_become", "ansible_become_user",
     }
 
     group_vars_dir = base / "group_vars"
@@ -411,6 +412,8 @@ def _host_from_vars(host_name: str, host_data: dict[str, Any]) -> HostConfig:
         "ansible_user",
         "ansible_connection",
         "ansible_python_interpreter",
+        "ansible_become",
+        "ansible_become_user",
     }
 
     return HostConfig(
@@ -422,6 +425,8 @@ def _host_from_vars(host_name: str, host_data: dict[str, Any]) -> HostConfig:
         ansible_python_interpreter=host_data.get(
             "ansible_python_interpreter", "python3"
         ),
+        ansible_become=host_data.get("ansible_become", False),
+        ansible_become_user=host_data.get("ansible_become_user", "root"),
         vars={k: v for k, v in host_data.items() if k not in standard_fields},
     )
 
--- a/src/ftl2/inventory.py
+++ b/src/ftl2/inventory.py
@@ -101,6 +101,16 @@ class Inventory:
 
 _VALID_VARS_EXTENSIONS = {".yml", ".yaml", ".json", ""}
 
+_STANDARD_HOST_FIELDS = {
+    "ansible_host",
+    "ansible_port",
+    "ansible_user",
+    "ansible_connection",
+    "ansible_python_interpreter",
+    "ansible_become",
+    "ansible_become_user",
+}
+
 
 def _load_vars_file(path: Path) -> dict[str, Any]:
     content = path.read_text()
@@ -142,12 +152,6 @@ def _vars_entry_name(entry: Path) -> str:
 def _apply_external_vars(inventory: Inventory, inventory_path: Path) -> None:
     """Load group_vars/ and host_vars/ directories adjacent to the inventory file."""
     base = inventory_path.parent
-    standard_fields = {
-        "ansible_host", "ansible_port", "ansible_user",
-        "ansible_connection", "ansible_python_interpreter",
-        "ansible_become", "ansible_become_user",
-    }
-
     group_vars_dir = base / "group_vars"
     if group_vars_dir.is_dir():
         for entry in sorted(group_vars_dir.iterdir()):
@@ -169,7 +173,7 @@ def _apply_external_vars(inventory: Inventory, inventory_path: Path) -> None:
             if host is None:
                 continue
             loaded = _load_vars_dir(entry)
-            for field_name in standard_fields:
+            for field_name in _STANDARD_HOST_FIELDS:
                 if field_name in loaded:
                     setattr(host, field_name, loaded.pop(field_name))
             host.vars.update(loaded)
@@ -406,7 +410,7 @@ def _host_from_vars(host_name: str, host_data: dict[str, Any]) -> HostConfig:
     Returns:
         HostConfig with standard fields extracted and remainder in vars
     """
-    standard_fields = {
+    _STANDARD_HOST_FIELDS = {
         "ansible_host",
         "ansible_port",
         "ansible_user",
@@ -427,7 +431,7 @@ def _host_from_vars(host_name: str, host_data: dict[str, Any]) -> HostConfig:
         ),
         ansible_become=host_data.get("ansible_become", False),
         ansible_become_user=host_data.get("ansible_become_user", "root"),
-        vars={k: v for k, v in host_data.items() if k not in standard_fields},
+        vars={k: v for k, v in host_data.items() if k not in _STANDARD_HOST_FIELDS},
     )
 
 
--- a/tests/test_inventory.py
+++ b/tests/test_inventory.py
@@ -1139,6 +1139,38 @@ class TestExternalVars:
         inv = load_inventory(inv_file)
         assert inv.get_group("webservers") is not None
 
+    def test_fqdn_host_vars(self, tmp_path):
+        """FQDN hostnames like db.example.com must not be truncated by Path.stem."""
+        inv_dir = tmp_path / "inventory"
+        inv_dir.mkdir()
+        (inv_dir / "hosts.yml").write_text(
+            "all:\n  hosts:\n    db.example.com:\n      ansible_host: 10.0.0.1\n"
+        )
+        hv = inv_dir / "host_vars"
+        hv.mkdir()
+        # Directory form — name is the full FQDN
+        fqdn_dir = hv / "db.example.com"
+        fqdn_dir.mkdir()
+        (fqdn_dir / "settings.yml").write_text("db_port: 5432\n")
+
+        inv = load_inventory(inv_dir / "hosts.yml")
+        host = inv.get_all_hosts()["db.example.com"]
+        assert host.vars["db_port"] == 5432
+
+    def test_fqdn_group_vars(self, tmp_path):
+        """Group names with dots in extensionless files must not be truncated."""
+        inv_dir = tmp_path / "inventory"
+        inv_dir.mkdir()
+        (inv_dir / "hosts.yml").write_text(
+            "us.east:\n  hosts:\n    web01:\n      ansible_host: 10.0.0.1\n"
+        )
+        gv = inv_dir / "group_vars"
+        gv.mkdir()
+        (gv / "us.east").write_text("region: us-east-1\n")
+
+        inv = load_inventory(inv_dir / "hosts.yml")
+        assert inv.get_group("us.east").vars["region"] == "us-east-1"
+
     def test_host_vars_promotes_standard_fields(self, tmp_path):
         inv_dir = tmp_path / "inventory"
         inv_dir.mkdir()
--- a/tests/test_inventory.py
+++ b/tests/test_inventory.py
@@ -1186,3 +1186,22 @@ class TestExternalVars:
         assert host.ansible_user == "deploy"
         assert host.vars["custom_var"] == "hello"
         assert "ansible_user" not in host.vars
+
+    def test_host_vars_promotes_become_fields(self, tmp_path):
+        inv_dir = tmp_path / "inventory"
+        inv_dir.mkdir()
+        (inv_dir / "hosts.yml").write_text(
+            "all:\n  hosts:\n    myhost:\n      ansible_host: 10.0.0.1\n"
+        )
+        hv = inv_dir / "host_vars"
+        hv.mkdir()
+        (hv / "myhost.yml").write_text(
+            "ansible_become: true\nansible_become_user: admin\n"
+        )
+
+        inv = load_inventory(inv_dir / "hosts.yml")
+        host = inv.get_group("all").get_host("myhost")
+        assert host.ansible_become is True
+        assert host.ansible_become_user == "admin"
+        assert "ansible_become" not in host.vars
+        assert "ansible_become_user" not in host.vars
--- a/tests/test_inventory.py
+++ b/tests/test_inventory.py
@@ -1007,3 +1007,150 @@ host1 flag= port=80
         h = inv.get_group("web").get_host("host1")
         assert h.vars["flag"] == ""
         assert h.vars["port"] == 80
+
+
+class TestExternalVars:
+    """Tests for host_vars/ and group_vars/ loading alongside inventory."""
+
+    def test_group_vars_yml(self, tmp_path):
+        inv_dir = tmp_path / "inventory"
+        inv_dir.mkdir()
+        (inv_dir / "hosts.yml").write_text(
+            "webservers:\n  hosts:\n    web01:\n      ansible_host: 10.0.0.1\n"
+        )
+        gv = inv_dir / "group_vars"
+        gv.mkdir()
+        (gv / "webservers.yml").write_text("http_port: 80\nmax_clients: 200\n")
+
+        inv = load_inventory(inv_dir / "hosts.yml")
+        ws = inv.get_group("webservers")
+        assert ws.vars["http_port"] == 80
+        assert ws.vars["max_clients"] == 200
+
+    def test_host_vars_yml(self, tmp_path):
+        inv_dir = tmp_path / "inventory"
+        inv_dir.mkdir()
+        (inv_dir / "hosts.yml").write_text(
+            "all:\n  hosts:\n    foosball:\n      ansible_host: 10.0.0.2\n"
+        )
+        hv = inv_dir / "host_vars"
+        hv.mkdir()
+        (hv / "foosball.yml").write_text("app_port: 9000\nenv: production\n")
+
+        inv = load_inventory(inv_dir / "hosts.yml")
+        host = inv.get_group("all").get_host("foosball")
+        assert host.vars["app_port"] == 9000
+        assert host.vars["env"] == "production"
+
+    def test_group_vars_directory(self, tmp_path):
+        inv_dir = tmp_path / "inventory"
+        inv_dir.mkdir()
+        (inv_dir / "hosts.yml").write_text(
+            "raleigh:\n  hosts:\n    host3:\n      ansible_host: 10.0.0.3\n"
+        )
+        gv = inv_dir / "group_vars" / "raleigh"
+        gv.mkdir(parents=True)
+        (gv / "a_db_settings.yml").write_text("db_port: 5432\n")
+        (gv / "b_cluster_settings.yml").write_text("cluster_name: east\ndb_port: 5433\n")
+
+        inv = load_inventory(inv_dir / "hosts.yml")
+        raleigh = inv.get_group("raleigh")
+        assert raleigh.vars["cluster_name"] == "east"
+        assert raleigh.vars["db_port"] == 5433  # b overrides a (sorted order)
+
+    def test_group_vars_all(self, tmp_path):
+        inv_dir = tmp_path / "inventory"
+        inv_dir.mkdir()
+        (inv_dir / "hosts.yml").write_text(
+            "webservers:\n  hosts:\n    web01:\n      ansible_host: 10.0.0.1\n"
+        )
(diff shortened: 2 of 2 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#115
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/inventory.py:_load_vars_file
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean helper. Uses `yaml.safe_load` (safe), handles JSON via suffix check. The `or {}` guard handles `None` from empty files. Called by `_load_vars_dir`.

### src/ftl2/inventory.py:_load_vars_dir
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly handles both file and directory cases. The `is_dir()` guard (present in on-disk code but missing from the submitted diff) prevents errors on broken symlinks or special files. Pre-filters invalid extensions before recursing. Sorted iteration ensures deterministic merge order, matching Ansible's behavior. Recursive design supports nested directory structures.

### src/ftl2/inventory.py:_apply_external_vars
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Two concerns:

1. **`standard_fields` is missing `ansible_become` and `ansible_become_user`**: `HostConfig` has `ansible_become` and `ansible_become_user` attributes that feed `become_config`. If a user sets `ansible_become: true` in `host_vars/myhost.yml`, it will land in `host.vars["ansible_become"]` instead of on the dataclass attribute, and `host.become_config.effective` will remain `False`. This is *consistent* with `_host_from_vars` (which has the same omission), but both are wrong. Since external vars are a more common place to set become config than inline inventory, this is worth flagging now.

2. **`entry.stem` with compound extensions**: A file like `host_vars/web01.backup.yml` would produce `host_name = "web01.backup"`, which won't match any host. This is arguably correct behavior (don't load unexpected files), but it differs from Ansible which uses the full directory/file name matching logic. Minor issue.

3. **Untested**: No test for `host_vars` as a directory (subdirectory with multiple files), only as single `.yml` files. The group_vars directory case is tested but the analogous host_vars directory case is not.

### src/ftl2/inventory.py:load_inventory (integration)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All four inventory format paths (script, JSON, INI, YAML) now call `_apply_external_vars(inv, path)` before returning. The pattern is consistent: load inventory, apply external vars, return. The `path` argument is always the original inventory file path, so `path.parent` correctly resolves to the directory containing the inventory file. The `test_ini_inventory_with_vars_dirs` test verifies cross-format integration.

### tests/test_inventory.py:TestExternalVars
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Good coverage of the happy paths: group_vars file, host_vars file, directory merging, "all" group auto-creation, JSON vars, no-extension files, invalid extension filtering, override semantics, no-op when dirs don't exist, INI format, and standard field promotion. Missing coverage for:
- **host_vars as a directory** (subdirectory with multiple files)
- **JSON inventory format with external vars**  
- **`.yaml` extension** (only `.yml` and `.json` tested)
- **Empty vars file** (would `yaml.safe_load("")` return `None`? The `or {}` guard should handle it, but untested)
- **Unrecognized host in host_vars/** (silently skipped — worth a test to document the behavior)

### Self-Review
**Limitations:** Could not run the test suite (no pytest available without `uv run` approval). The submitted diff differs from the on-disk code — the on-disk version is improved (has `is_dir()` guard, pre-filters extensions, cleaner variable naming). Reviewed both but focused on the on-disk version. Could not verify whether other callers of `load_inventory` (e.g., `AutomationContext`) would need changes for the new external vars behavior.

### Feature Requests
- Include the actual on-disk diff (from `git diff`) rather than a manually-provided diff when they diverge — I had to discover the discrepancy myself
- Show callers of modified public functions to verify integration impact

## gemini [CONCERN]

### src/ftl2/inventory.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The use of `entry.stem` to extract the host name (and group name) is flawed for names containing dots, such as Fully Qualified Domain Names (FQDNs). In Python's `pathlib`, `Path("db.example.com").stem` evaluates to `'db.example'`, stripping the last component as if it were an extension. If a user creates a `host_vars/db.example.com/` directory or a `host_vars/db.example.com` file without an extension, the host name will be parsed incorrectly and the variables will be silently skipped. To fix this, you should only strip the suffix if it explicitly matches one of the known extensions (`.yml`, `.yaml`, `.json`), and use `entry.name` otherwise.

### tests/test_inventory.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The tests are comprehensive and cover various scenarios including extensions, directories, precedence, inline overrides, and standard field promotion. Adding a test case for an FQDN hostname (e.g., `db.example.com`) either as a directory or an extensionless file would have caught the `entry.stem` bug mentioned above.

### Self-Review
**Limitations:** None. The provided diff and context were sufficient to review the feature and identify the edge case with FQDN path stems.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#115
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### src/ftl2/inventory.py:_load_vars_file
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Simple dispatcher — JSON vs YAML. The `or {}` guards handle `None` from empty files correctly. Called by `_load_vars_dir` which is well-tested.

### src/ftl2/inventory.py:_load_vars_dir
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The recursive call at line 126 descends into subdirectories unconditionally (only *files* with bad extensions are skipped via `continue`). If a subdirectory contains a nested subdirectory, it will recurse into that too. Ansible limits vars directory loading to one level deep (files only, no recursive descent into nested subdirectories). This is unlikely to cause real-world breakage but is a behavioral divergence. Additionally, there's no test for deeply nested subdirectory structures.

### src/ftl2/inventory.py:_vars_entry_name
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly handles FQDN names by only stripping known extensions. `Path("db.example.com").suffix` returns `".com"` which is not in the set, so `entry.name` is returned — preserving the full FQDN. Tested by `test_fqdn_host_vars` and `test_fqdn_group_vars`.

### src/ftl2/inventory.py:_apply_external_vars
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Two issues: (1) The `standard_fields` set at line 145 includes `ansible_port` but HostConfig declares it as `int`. If a YAML file has `ansible_port: "2222"` (string), `setattr` will set a string where an int is expected — no type coercion. The `_host_from_vars` function doesn't have this problem because it uses `.get()` with defaults, but that function also doesn't coerce types, so it's a pre-existing issue. (2) `standard_fields` is missing `ansible_become` and `ansible_become_user`, so those will end up in `host.vars` rather than being promoted to dataclass fields. This is inconsistent with `_host_from_vars` which also doesn't handle those (pre-existing), but worth noting.

### src/ftl2/inventory.py:load_inventory (modified)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All four code paths (script, JSON, INI, YAML) now call `_apply_external_vars(inv, path)` before returning. Clean integration pattern — capture the result, apply external vars, return. No code path is missed.

### tests/test_inventory.py:TestExternalVars
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test suite covering: YAML group_vars file, host_vars file, directory-based group_vars with sorted merge order, `all` group auto-creation, JSON format, extensionless files, invalid extension filtering, override semantics, no-vars-dirs graceful handling, FQDN names for both hosts and groups, and standard field promotion. All tests use `tmp_path` fixtures correctly and exercise through the public `load_inventory` entry point.

### Self-Review
**Limitations:** Could not verify how `_apply_external_vars` interacts with the `AutomationContext` at runtime (whether group_vars are properly merged with host-level vars during module execution). Could not verify Ansible's actual behavior for recursive subdirectory loading to confirm the divergence claim.

### Feature Requests
- Show the full function body (not just diff hunks) for all modified functions and their callers, to verify integration more confidently
- Include the related issue body when the branch name references one (e.g., issue-108)

## gemini [BLOCK]

### ERROR
**Verdict:** BLOCK

Model invocation failed: Model gemini timed out after 300s

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#115
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/inventory.py:_load_vars_file
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Simple dispatch between JSON and YAML parsing. Handles empty files correctly via `or {}`. Consistent with existing YAML/JSON parsing patterns in the file.

### src/ftl2/inventory.py:_load_vars_dir
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The extension filtering logic only applies inside subdirectories, not at the top level of `group_vars/` or `host_vars/`. At the top level in `_apply_external_vars`, every entry is passed to `_load_vars_dir` regardless of extension. A file like `group_vars/README.md` would be attempted as YAML (the group name `README.md` wouldn't match any group, so it's harmlessly skipped — but it would still be parsed). More importantly, there's no protection against recursive symlink loops — `sorted(dirpath.iterdir())` followed by recursive `_load_vars_dir` on child directories could infinite-loop on a symlink cycle.

### src/ftl2/inventory.py:_vars_entry_name
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly preserves FQDN hostnames like `db.example.com` (whose suffix `.com` is not in the valid set) while stripping `.yml`/`.yaml`/`.json`. The logic matches the test cases for both FQDN hosts and dotted group names.

### src/ftl2/inventory.py:_apply_external_vars
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Two concerns: (1) `get_all_hosts()` returns only the first HostConfig instance when a host appears in multiple groups (since each group creates a separate `HostConfig` via `_host_from_vars`). External vars and standard field promotion only apply to one instance, leaving other group copies unmodified. This could lead to inconsistencies when accessing a host through different groups. (2) `standard_fields` is defined locally and must be kept in sync with `_host_from_vars` — a maintenance risk. There's no test for `ansible_become` or `ansible_become_user` promotion via host_vars.

### src/ftl2/inventory.py:load_inventory (modification)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Clean integration — `_apply_external_vars` is called at every exit point of `load_inventory`, applied consistently across all four format paths (script, JSON, INI, YAML). The refactoring from direct returns to `inv = ...; _apply_external_vars(inv, path); return inv` is correct.

### src/ftl2/inventory.py:_host_from_vars (modification)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

`ansible_become` and `ansible_become_user` are correctly added to `standard_fields` and extracted with appropriate defaults matching `HostConfig` (verified: `types.py:134-135` has `ansible_become: bool = False` and `ansible_become_user: str = "root"`). These fields are filtered out of `vars` dict. However, there's no dedicated test for these new fields being parsed from inline inventory data.

### tests/test_inventory.py:TestExternalVars
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

13 well-structured tests covering: file formats (YAML, JSON, no extension), directory-based vars, sorted override order, extension filtering, FQDN preservation, `all` group auto-creation, inline override, no-vars-dirs no-op, and standard field promotion. Missing coverage: (1) No test for INI inventory + external vars, (2) no test for `ansible_become`/`ansible_become_user` promotion, (3) no test for a host in multiple groups receiving external vars, (4) no test for empty vars files, (5) no test for the executable script path.

### Self-Review
**Limitations:** Could not see how HostConfig objects are accessed at execution time (through AutomationContext/ModuleProxy) to assess the practical impact of the multi-group host identity concern. Could not run the tests to verify they pass. Did not see the issue referenced in the branch name (issue #108) to verify intent.

### Feature Requests
- Include the linked issue body when the branch name contains an issue reference
- Show callers of modified functions (e.g., who calls `get_all_hosts()` and how hosts are resolved at execution time)

## gemini [PASS]

### src/ftl2/inventory.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The implementation correctly resolves and recursively loads `group_vars` and `host_vars`, handling directory iterations in sorted lexicographical order as expected in an Ansible-like inventory engine. It securely targets files via extension allowlisting while skillfully retaining FQDN hostnames using intelligent suffix parsing logic. Integration is complete, as it cleanly applies after initial inventory file loading for all parsed formats (YAML, INI, JSON, and Scripts). Variable precedence is handled logically (host vars overwrite group vars and inline definitions).

### tests/test_inventory.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The test suite comprehensively addresses the new features. It tests inline precedence, FQDN logic, standard field promotion, specific extension ignorances, and formats (JSON, YAML). Coverage is exceptional.

### Self-Review
**Limitations:** No specifications, issue descriptions, or architecture belief documents were provided to check against broader design goals, though the implementation cleanly meets typical expectations for this feature type.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#115
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### `src/ftl2/inventory.py:_load_vars_file`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Handles JSON via `json.loads` and YAML via `yaml.safe_load`, both with `or {}` to guard against empty files returning `None`. Uses `safe_load` (no security concern).

### `src/ftl2/inventory.py:_load_vars_dir`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The recursive directory traversal works correctly for the happy path. However, extensionless files (suffix `""`) are in `_VALID_VARS_EXTENSIONS`, which means hidden files like `.DS_Store` inside a vars subdirectory (e.g. `group_vars/webservers/.DS_Store`) would pass the extension filter and be loaded as YAML, likely causing a `UnicodeDecodeError` or YAML parse error. Ansible itself skips hidden files (names starting with `.`). Adding an `entry.name.startswith(".")` check would match Ansible's behavior and prevent macOS-specific crashes. No test covers hidden files inside vars subdirectories.

### `src/ftl2/inventory.py:_vars_entry_name`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly distinguishes FQDN names from files with known extensions. `Path("db.example.com").suffix` returns `.com` which is not in the valid set, so `entry.name` preserves the full FQDN. `Path("webservers.yml").suffix` returns `.yml` so `entry.stem` strips it. Well-tested with `test_fqdn_host_vars` and `test_fqdn_group_vars`.

### `src/ftl2/inventory.py:_apply_external_vars`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Two concerns: (1) The `standard_fields` set is duplicated between here (line 145) and `_host_from_vars` (line 409). If a field is added to one but not the other, behavior diverges silently. Consider extracting a module-level constant. (2) `get_all_hosts()` at line 165 returns only the first `HostConfig` instance found for each name. If a host appears in multiple groups with separate `HostConfig` instances (which happens with the current YAML loader), host_vars are only applied to the first instance. The other copies in other groups retain stale values. This matches a pre-existing design issue in `get_all_hosts()` but the host_vars feature makes it more visible. Group_vars handling is clean — the "all" group creation for `group_vars/all` is a nice touch.

### `src/ftl2/inventory.py:_host_from_vars` (changes)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Adds `ansible_become` and `ansible_become_user` to both the `standard_fields` set and the `HostConfig` constructor call with correct defaults (`False` and `"root"` respectively), matching the dataclass definition in `types.py:134-135`. This ensures these fields are promoted from `vars` dict to typed attributes when loading from inline inventory data.

### `src/ftl2/inventory.py:load_inventory` (changes)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All four inventory format paths (script, JSON, INI, YAML) consistently apply `_apply_external_vars` after loading. The pattern of `inv = load_*(); _apply_external_vars(inv, path); return inv` is clean and uniform. External vars are applied after the base inventory is built, so they correctly override inline vars (tested by `test_external_vars_override_inline`).

### `tests/test_inventory.py:TestExternalVars`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

14 tests covering: YAML/JSON/extensionless file formats, directory-form group vars with sorted merge order, "all" group auto-creation, invalid extension filtering, override semantics, no-vars-dir graceful handling, FQDN preservation for both hosts and groups, and standard field promotion including become fields. Missing coverage: INI and JSON inventory formats with external vars (all tests use YAML inventory files), and hidden files in vars subdirectories. The covered scenarios are well-chosen and reflect real-world Ansible usage patterns.

### Self-Review
**Limitations:** Could not verify behavior of `_apply_external_vars` when called with a non-file-backed inventory (e.g. inventory loaded from string content where `path.parent` might not exist meaningfully — though all code paths in `load_inventory` start from a real file path so this is safe). Could not verify whether hosts in multiple groups share the same `HostConfig` instance or have separate copies.

### Feature Requests
- Show callers of modified functions (e.g. who calls `_host_from_vars`) to verify integration completeness
- Include the HostConfig dataclass definition automatically when inventory.py is in the diff, since they're tightly coupled

## gemini [PASS]

### src/ftl2/inventory.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

The implementation for loading `group_vars` and `host_vars` is robust and correct. It handles files, directories, FQDNs without truncating them improperly, and sorts directory entries as expected for precedence. It correctly promotes standard fields like `ansible_become` and `ansible_become_user` to the `HostConfig` object. The integration is fully wired into all inventory loaders (script, JSON, INI, YAML). Note: An empty `.json` file in `group_vars` will raise a `json.decoder.JSONDecodeError` rather than returning `{}` since `json.loads("")` is invalid JSON, but this is standard standard library behavior and acceptable.

### tests/test_inventory.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

The test coverage is comprehensive. It thoroughly tests YAML/JSON files, extensionless files, FQDN naming, directory precedence overriding, ignoring invalid file extensions, and field promotion on the host objects.

### Self-Review
**Limitations:** I do not have access to the `HostConfig` and `HostGroup` class implementations to strictly verify that `ansible_become` fields are valid dataclass arguments, or how the `all` group inheritance works. However, the comprehensive passing tests strongly imply these integrations are structurally correct.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#115
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### `src/ftl2/inventory.py:_STANDARD_HOST_FIELDS` (module-level constant)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The module-level `_STANDARD_HOST_FIELDS` set is defined at line 104, but `_host_from_vars` at line 413 shadows it with a local variable of the same name. This is confusing — the local re-declaration is identical in content but uses a misleading name (uppercase convention implies module constant). It works correctly since both have the same members, but the shadowing is a maintenance trap: updating one without the other would cause silent divergence. The local should either be removed in favor of the module-level constant, or renamed to `standard_fields` (the original local name before this diff).

### `src/ftl2/inventory.py:_load_vars_file`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

No error handling for malformed files — `yaml.safe_load` or `json.loads` will raise uncaught exceptions. This is acceptable at a system boundary (external files), but for extensionless files (suffix `""`), `_load_vars_file` parses them as YAML. That's correct for Ansible compatibility but worth noting. More importantly, the `""` extension in `_VALID_VARS_EXTENSIONS` means extensionless files pass through to `_load_vars_file`, which treats them as YAML. No test covers an extensionless file *inside* a directory (the `test_no_extension_vars_file` test covers a top-level extensionless file, not a nested one).

### `src/ftl2/inventory.py:_load_vars_dir`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The recursive traversal has no depth limit. A symlink loop or deeply nested directory structure could cause infinite recursion. Ansible itself limits directory depth. Also, line 134 checks `entry.is_file()` before filtering by extension, but `entry.is_dir()` entries pass through unconditionally and recurse — this is intentional (directories are always recursed into) but means a subdirectory named `readme.txt` would be recursed into rather than skipped. This is an edge case but could be surprising.

### `src/ftl2/inventory.py:_vars_entry_name`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly handles the FQDN case — `db.example.com` as an extensionless file or directory returns the full name, while `db.example.com.yml` would return `db.example`. This matches Ansible's behavior. Well-tested with dedicated FQDN tests.

### `src/ftl2/inventory.py:_apply_external_vars`
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Two issues: (1) `get_all_hosts()` at line 169 returns a *copy* of the hosts dict, but the `HostConfig` objects within it are shared references — so `setattr(host, ...)` at line 178 correctly mutates the original host. This works but is fragile; if `get_all_hosts` ever deep-copies, this breaks silently. (2) For `group_vars/all`, the code creates a new `HostGroup(name="all")` if one doesn't exist (line 161-162), which is correct. However, for host_vars, hosts that exist in the inventory but whose name doesn't match a `host_vars/` entry are silently skipped (line 173-174) — this is correct Ansible behavior. Missing test: host_vars for a host that appears in multiple groups (verifying the mutation affects the shared object).

### `src/ftl2/inventory.py:_host_from_vars` (local `_STANDARD_HOST_FIELDS` shadow)
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The function correctly adds `ansible_become` and `ansible_become_user` to HostConfig construction. However, as noted above, it shadows the module-level `_STANDARD_HOST_FIELDS` with an identical local. The original code used a local named `standard_fields`; the rename to `_STANDARD_HOST_FIELDS` creates a confusing pattern where the same constant-style name is defined twice in the module. This should use the module-level constant directly — `from` line 104 — instead of redeclaring it.

### `src/ftl2/inventory.py:load_inventory` (integration of `_apply_external_vars`)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All four inventory format paths (script, JSON, INI, YAML) now apply external vars before returning. The pattern is consistent. External vars are applied after the inventory is fully loaded, which is the correct ordering — inline vars exist first, then external vars overlay them via `.update()`.

### `tests/test_inventory.py:TestExternalVars`
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test class covering: group_vars (yml, json, extensionless, directory), host_vars (yml, FQDN), "all" group creation, invalid extension filtering, override semantics, no-vars-dirs baseline, standard field promotion, and become field promotion. Good edge case coverage for FQDN with dots. Missing: INI/JSON inventory formats with external vars (all tests use YAML inventory), host appearing in multiple groups, empty vars files, and nested subdirectories within a vars directory.

### Self-Review
**Limitations:** Could not verify whether `get_all_hosts()` returns shared references vs copies of `HostConfig` without reading the full dataclass definition (confirmed it does share references by reading the method). Did not verify INI inventory loader's interaction with external vars at runtime. Could not check if any other callers of `_host_from_vars` or `_STANDARD_HOST_FIELDS` exist outside this file.

### Feature Requests
- Show all callers of modified/new functions (e.g., who calls `_host_from_vars`, `_apply_external_vars`) to verify integration completeness
- Flag duplicated constants/variables automatically (the `_STANDARD_HOST_FIELDS` shadowing is a real bug vector)
- Include related test files for all modified source files automatically, not just when prefixed

## gemini [PASS]

### src/ftl2/inventory.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The implementation effectively adds support for loading `group_vars` and `host_vars`, accurately mirroring standard Ansible behavior and priorities (external variables correctly override inline inventory variables). FQDN hostnames, nested variables directories, and varying extensions are handled well.

Two minor observations that don't impact correctness but could be improved:
1. `_STANDARD_HOST_FIELDS` is defined twice (once at the module level and once redundantly inside `_host_from_vars`). This is a minor DRY violation.
2. `_load_vars_file` assumes `yaml.safe_load` always returns a dictionary (`res or {}`). If a user maliciously or accidentally provides a file containing a top-level string or list (e.g., `echo "just a string" > group_vars/webservers/README`), the subsequent `.update(loaded)` will raise a `ValueError` or `TypeError`. Explicit type checking could provide a cleaner error message, but crashing on invalid config shapes is generally acceptable.

### tests/test_inventory.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The new test suite `TestExternalVars` is robust and comprehensive. It accurately covers JSON/YAML format parsing, ignored extensions, FQDN extraction integrity, inline overriding behaviors, and promotion of `ansible_become` fields.

### Self-Review
**Limitations:** None. The provided diff and code context were sufficient to evaluate correctness, test coverage, and functionality. Ensure that `HostConfig` explicitly defines or accepts `ansible_become` and `ansible_become_user` fields since they are being applied via `setattr()` and passed to the constructor.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#115
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/inventory.py:_STANDARD_HOST_FIELDS
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Consolidating the formerly local `standard_fields` set into a module-level constant and adding `ansible_become`/`ansible_become_user` is correct. The constant is used both in `_host_from_vars` (filtering vars dict) and `_apply_external_vars` (promoting fields via setattr). The fields match the `HostConfig` dataclass attributes exactly.

### src/ftl2/inventory.py:_load_vars_file
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Straightforward file reader dispatching on `.json` suffix vs YAML. The `or {}` guard handles `None` returns from `yaml.safe_load` and `json.loads` correctly (empty file or `null` content). Extensionless files fall through to the YAML path, which is correct since YAML is a superset and extensionless files are treated as YAML by Ansible.

### src/ftl2/inventory.py:_load_vars_dir
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The recursive walk is correct for normal cases, but has an edge case: if a directory entry inside a vars subdirectory is itself a directory (nested subdirectory), the `entry.suffix not in _VALID_VARS_EXTENSIONS` check is skipped (it only applies to `entry.is_file()`), so the recursion will descend into arbitrary subdirectories. This matches Ansible's behavior of recursing into subdirectories, so it's not a bug per se, but there's no test for nested subdirectories inside a vars dir. Also, if a vars file contains invalid YAML/JSON, the exception propagates with no context about which file failed — acceptable but worth noting.

### src/ftl2/inventory.py:_vars_entry_name
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly handles the FQDN edge case. `Path("db.example.com").suffix` is `".com"` which is not in the valid extensions list, so it returns `entry.name` = `"db.example.com"`. For `"webservers.yml"`, suffix is `".yml"`, so it returns `entry.stem` = `"webservers"`. This is sound. The tests at `test_fqdn_host_vars` and `test_fqdn_group_vars` cover the critical FQDN case.

### src/ftl2/inventory.py:_apply_external_vars
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Two concerns: (1) `get_all_hosts()` returns a *new dict* with references to existing `HostConfig` objects. The `setattr` mutations on those objects work because `HostConfig` is a mutable dataclass — the mutation propagates back to the inventory's groups. This is correct but subtle and fragile (if `HostConfig` were ever made frozen, this would break silently at runtime). (2) If a host appears in multiple groups, `get_all_hosts()` returns only the first occurrence (line 98: `if host_name not in result`). The `setattr` call mutates that object, and if all groups share the same `HostConfig` instance this is fine. But if groups have distinct `HostConfig` objects for the same host name, only the first one gets the host_vars applied. This is likely fine for real-world usage since hosts are typically the same object, but there's no test verifying this edge case. (3) The `loaded.pop(field_name)` mutates the loaded dict to avoid duplicate keys in `host.vars.update()` — clean approach.

### src/ftl2/inventory.py:load_inventory (integration wiring)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All four inventory format paths (script, JSON, INI, YAML) now call `_apply_external_vars(inv, path)` before returning. The `path` variable is always a `Path` at this point, so `path.parent` gives the correct directory for locating `group_vars/` and `host_vars/`. Integration is complete.

### src/ftl2/inventory.py:_host_from_vars
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Now uses the module-level `_STANDARD_HOST_FIELDS` instead of a local set, and adds `ansible_become`/`ansible_become_user` extraction. The defaults (`False` and `"root"`) match `HostConfig`'s defaults. The formerly shadowed local variable is gone (commit `ee45053`).

### tests/test_inventory.py:TestExternalVars
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Comprehensive test suite covering: yml/json/extensionless group_vars files, directory-form group_vars with sorted merge, the `all` group auto-creation, invalid extension filtering, override semantics (external > inline), no-op when dirs don't exist, FQDN hostnames (both host_vars and group_vars), and standard field promotion (including become fields). All tests use `tmp_path` and `load_inventory` for proper integration testing. Missing: host_vars directory form (subdirectory per host with multiple files), INI/JSON inventory format tests with external vars, and edge case where host_vars references a host not in inventory (implicitly covered by the `continue` in the code).

### Self-Review
**Limitations:** Could not verify whether any other callers of `_host_from_vars` or `_STANDARD_HOST_FIELDS` exist beyond inventory.py. Did not run the test suite to confirm all tests pass. Could not verify the `get_all_hosts()` object identity guarantee (whether hosts in multiple groups share the same `HostConfig` instance or not).

### Feature Requests
- Include the output of `git log --oneline` for the branch to see all commits in context, not just the diff
- Show callers of modified functions (e.g., who calls `_host_from_vars`) to verify integration completeness

## gemini [PASS]

### src/ftl2/inventory.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The implementation correctly and robustly adds support for `group_vars` and `host_vars` sidecar directories. It safely recursively evaluates variables, respects typical Ansible extension filtering (`.yml`, `.yaml`, `.json`, and extensionless), and properly merges standard host fields onto the `HostConfig` objects. Existing module imports (`json`, `yaml`, `Path`) are verified to already be present given the context of the diff.

### tests/test_inventory.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Excellent test coverage. The new tests use standard `pytest` fixtures (`tmp_path`) and successfully evaluate YAML/JSON loads, FQDN extraction logic, directory ordering, extension exclusion, inline vars overrides, and special case handling for `ansible_become` and `ansible_become_user`.

### Self-Review
**Limitations:** No specifications or issues were provided. The full `src/ftl2/inventory.py` was not accessible, requiring assumptions that standard classes like `HostConfig` correctly support fields like `ansible_become` as strongly implied by the usage in the `_host_from_vars` return block.

PR #121, opened 2026-04-25
Bundle events.py in gate for remote FTL module support

### LINKED ISSUE #120: FTL modules fail on remote hosts: ftl2.events and ftl2.ftl_modules.exceptions not available in gate
## Problem

FTL modules like `ftl2/ftl_modules/http.py` (which provides `get_url` and `uri`) import from `ftl2.events` and `ftl2.ftl_modules.exceptions`. These modules work fine locally but fail when executed on remote hosts via the gate because the gate package does not include these `ftl2` internals.

```
FTL module execution failed: No module named 'ftl2.events'
```

## Root Cause

`http.py` imports:
```python
from ftl2.events import emit_progress
from ftl2.ftl_modules.exceptions import FTLModuleError
```

The gate `.pyz` bundles module files and their Ansible `module_utils` dependencies, but not arbitrary `ftl2` package modules. When the gate runs `http.py` on a remote host, these imports fail.

## Analysis

Both `ftl2.events` and `ftl2.ftl_modules.exceptions` are stdlib-only -- they have zero transitive dependencies on the rest of `ftl2`. So they could be bundled, but the current approach of importing directly from `ftl2` creates an unclear boundary between local-only and gate-portable code.

Ansible solved this with `module_utils` -- a flat namespace specifically designed to be shipped alongside modules to remote hosts.

## Proposed Fix

Create `ftl2/ftl_modules/module_utils/` for code that FTL modules need both locally and remotely. Move `events` and `exceptions` (or the parts needed by modules) into `module_utils/`, and update the gate builder to bundle `module_utils/` automatically -- the same way it already bundles Ansible `module_utils` dependencies.

This establishes the rule: anything in `ftl_modules/` (including its `module_utils/`) can run remotely; everything else in `ftl2/` is local-only.

## Workaround

Currently `ftl2-servercraft` works around this by using `ftl.minecraft.shell(cmd="curl ...")` instead of `ftl.minecraft.get_url(...)` for remote downloads.

## Files

- `src/ftl2/ftl_modules/http.py` -- the failing module
- `src/ftl2/events.py` -- stdlib-only, needs to be gate-portable
- `src/ftl2/ftl_modules/exceptions.py` -- stdlib-only, needs to be gate-portable
- `src/ftl2/gate.py` -- gate builder, needs to bundle module_utils

benthomasson:
## Summary

Fixes #120 — FTL modules like http.py fail on remote hosts because the gate doesn't include ftl2.events.

## Changes

- `src/ftl2/gate.py` — 3 additions following the existing `_copy_ftl_module_exceptions()` pattern:
  - `events.py` added to `compute_hash()` source list for cache invalidation
  - `_copy_events_module()` call added in `_build_new_gate()`
  - New `_copy_events_module()` method with soft-failure behavior
- `tests/test_gate_events_bundling.py` — 10 new tests

## Test plan

- [x] 10 new tests pass (hash, bundling, content matching, import, soft failure)
- [x] 32 existing gate tests pass (no regressions)
- [x] Gate .pyz contains ftl2/events.py with emit_progress
- [x] Modules importing both events and exceptions work in gate runtime

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/ftl2/gate.py
+++ b/src/ftl2/gate.py
@@ -100,6 +100,7 @@ class GateBuildConfig:
                 ftl2_dir / "ftl_gate" / "__main__.py",
                 ftl2_dir / "message.py",
                 ftl2_dir / "ftl_modules" / "exceptions.py",
+                ftl2_dir / "events.py",
                 ftl2_dir / "policy.py",
                 ftl2_dir / "exceptions.py",
             ]:
@@ -229,6 +230,9 @@ class GateBuilder:
 
             # Copy FTL module exceptions (needed by FTL modules sent via FTLModule messages)
             self._copy_ftl_module_exceptions(ftl2_dir)
+
+            # Copy events module (needed by FTL modules that emit progress events)
+            self._copy_events_module(ftl2_dir)
 
             # Copy policy engine and its dependency (exceptions.py) for gate-side enforcement
             self._copy_policy_module(ftl2_dir)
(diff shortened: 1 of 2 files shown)


benthomasson:
## Code Review (Claude + Gemini)

**Gate: CONCERN** (one valid observation, not blocking)

### Claude [CONCERN]

**`gate.py` changes** — All PASS. The three additions (hash, call, method) follow the established pattern exactly. Consistent with beliefs `gate-hash-includes-runtime-source`, `gate-copy-exceptions-non-fatal`, `events-zero-external-deps`.

**`TestEventsInGateBuild`** — PASS. Strong tests: verifies inclusion in zipapp, byte-for-byte content match, and key symbols present. These are the tests that actually validate the fix.

**`TestFTLModuleWithEventsImport`** — CONCERN. These tests call `execute_ftl_module` with modules that import `ftl2.events`, but they run in the test process where `ftl2.events` is already importable via the installed package — not inside a gate zipapp. The imports succeed regardless of whether events.py is bundled. Useful as regression tests for the execution path, but they don't validate the actual gate bundling. The `TestEventsInGateBuild` tests cover that.

### Gemini [PASS]

All changes PASS. Straightforward, well-contained, adequately covered.

---

*Reviewed with [multi-model-code-review](https://github.com/benthomasson/multi-model-code-review) using ftl2-expert beliefs*

PR #142, opened 2026-05-28
Fix become_user HOME env var (sudo -H)

### LINKED ISSUE #141: become_user does not set HOME, breaks uv pip install
## Problem

When using `become_user` in a deploy script, the `HOME` environment variable is not updated to the target user's home directory. This causes tools like `uv` to look for config files in the wrong location.

### Example

```python
await ftl[HOST].shell(
    cmd='uv pip install --reinstall-package myapp /home/expert/myapp.whl',
    become_user='expert',
)
```

`uv` runs as the `expert` user but `HOME` remains `/home/admin`, so it tries to read `/home/admin/uv.toml` and fails with:

```
error: failed to open file `/home/admin/uv.toml`: Permission denied (os error 13)
```

### Workaround

Using `su - expert -c "..."` via `shell()` instead of `become_user` works correctly because `su -` sets up the full login environment including `HOME`.

### Expected behavior

`become_user` should set `HOME` to the target user's home directory (equivalent to `sudo -u expert -H` or `su - expert`), matching Ansible's `become_user` behavior.

### Impact

This silently breaks package installation — `uv pip install` appears to succeed but doesn't actually update files when the version number hasn't changed, because it can't read/write its cache properly.

benthomasson:
## Summary

- Adds `-H` flag to `sudo` invocations in `BecomeConfig.become_prefix()` so `HOME` is set to the target user's home directory
- Updates three test assertions in `test_become.py` to match

Fixes #141

## Test plan

- [x] `tests/test_become.py` -- all 30 tests pass
- [ ] Manual: `become_user="expert"` + `shell(cmd="echo $HOME")` returns `/home/expert`

### DIFF
--- a/src/ftl2/types.py
+++ b/src/ftl2/types.py
@@ -68,8 +68,8 @@ class BecomeConfig:
 
         if method == "sudo":
             if user == "root":
-                return f"sudo -n {cmd}"
-            return f"sudo -n -u {user} {cmd}"
+                return f"sudo -n -H {cmd}"
+            return f"sudo -n -H -u {user} {cmd}"
         elif method == "doas":
             if user == "root":
                 return f"doas -n {cmd}"
--- a/tests/test_become.py
+++ b/tests/test_become.py
@@ -28,16 +28,16 @@ class TestBecomeConfig:
 
     def test_become_prefix_root(self):
         bc = BecomeConfig(become=True)
-        assert bc.become_prefix("whoami") == "sudo -n whoami"
+        assert bc.become_prefix("whoami") == "sudo -n -H whoami"
 
     def test_become_prefix_nonroot_user(self):
         bc = BecomeConfig(become=True, become_user="catbeez")
-        assert bc.become_prefix("whoami") == "sudo -n -u catbeez whoami"
+        assert bc.become_prefix("whoami") == "sudo -n -H -u catbeez whoami"
 
     def test_become_prefix_preserves_command(self):
         bc = BecomeConfig(become=True)
         cmd = "/bin/sh -c 'firewall-cmd --reload'"
-        assert bc.become_prefix(cmd) == f"sudo -n {cmd}"
+        assert bc.become_prefix(cmd) == f"sudo -n -H {cmd}"
 
     def test_become_prefix_doas_root(self):
         bc = BecomeConfig(become=True, become_method="doas")
--- a/tests/test_become_edge_cases.py
+++ b/tests/test_become_edge_cases.py
@@ -12,7 +12,7 @@ class TestBecomeEdgeCases:
         """Deprecated sudo_prefix() still works and returns same result."""
         bc = BecomeConfig(become=True)
         assert bc.sudo_prefix("whoami") == bc.become_prefix("whoami")
-        assert bc.sudo_prefix("whoami") == "sudo -n whoami"
+        assert bc.sudo_prefix("whoami") == "sudo -n -H whoami"
 
     def test_su_single_quote_in_command(self):
         """su -c properly escapes commands containing single quotes via shlex.quote."""
@@ -44,7 +44,7 @@ class TestBecomeEdgeCases:
     def test_empty_command(self):
         """Empty command string doesn't crash."""
         bc = BecomeConfig(become=True)
-        assert bc.become_prefix("") == "sudo -n "
+        assert bc.become_prefix("") == "sudo -n -H "
 
     def test_doas_complex_command(self):
         """doas with a multi-part command."""


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#142
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [BLOCK]

### src/ftl2/types.py:become_prefix
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Adding `-H` to `sudo` is correct — it sets `HOME` to the target user's home directory, which is important for modules that depend on `~` resolving properly (e.g., pip, dnf cache). The flag is only added to the `sudo` path, not `doas` or `su`, which is appropriate since `-H` is sudo-specific. The single caller in `runners.py:1011` requires no changes since it just calls `become_prefix()` on the constructed command. The commit message ("Add -H to sudo in become_prefix so HOME is set to target user's home dir") matches exactly what was done.

### tests/test_become.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All three modified assertions correctly reflect the new `-H` flag. The `doas` and `su` tests are unchanged, which is correct since only `sudo` was modified.

### tests/test_become_edge_cases.py (NOT IN DIFF — BROKEN)
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** PARTIAL

This file was **not updated** but contains two assertions that will now fail:
- **Line 15**: `assert bc.sudo_prefix("whoami") == "sudo -n whoami"` — `sudo_prefix` delegates to `become_prefix`, so this now returns `"sudo -n -H whoami"`.
- **Line 47**: `assert bc.become_prefix("") == "sudo -n "` — now returns `"sudo -n -H "`.

These must be updated to include `-H` or the test suite will break.

### Self-Review
**Limitations:** Did not run `pytest` to confirm the test failures — assessment is based on code reading. Did not check whether `-H` interacts unexpectedly with `-n` on any sudo version, though this is a standard combination.

### Feature Requests
- Auto-include test files that import or reference modified symbols, even when they aren't in the diff — this would have caught the `test_become_edge_cases.py` breakage immediately.

## gemini [BLOCK]

### ERROR
**Verdict:** BLOCK

Model invocation failed: Model gemini failed: [31mGemini CLI is not running in a trusted directory. To proceed, either use `--skip-trust`, set the `GEMINI_CLI_TRUST_WORKSPACE=true` environment variable, or trust this directory in interactive mode. For more details, see https://geminicli.com/docs/cli/trusted-folders/#headless-and-automated-environments[0m

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#142
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [PASS]

### src/ftl2/types.py:become_prefix
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Adding `-H` to `sudo` sets `HOME` to the target user's home directory, fixing a real problem where `$HOME` remains the invoking user's home after privilege escalation. This matches Ansible's own default `sudo` behavior (`sudo -H`). The flag is correctly applied only to `sudo` — `doas` already sets HOME to the target user by default, and `su -` handles it via the `-` (login shell) flag. All callers in `runners.py:1011` and `proxy.py` (lines 435, 447, 450, 456, 457, 461, 465, 467, 469, 871) pass through `become_prefix()` and will pick up the change automatically with no signature change.

### tests/test_become.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

All three affected assertions (`test_become_prefix_root`, `test_become_prefix_nonroot_user`, `test_become_prefix_preserves_command`) are updated to expect `-H`. The `doas` and `su` tests are correctly left unchanged.

### tests/test_become_edge_cases.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Edge case assertions updated consistently: backward-compat `sudo_prefix()` test, empty command test. The `doas_complex_command` test is correctly unchanged since `-H` only applies to `sudo`.

### Self-Review
**Limitations:** Could not verify whether there are integration tests that exercise `become_prefix` end-to-end over SSH. Could not check issue #141 for full requirements. Did not verify whether any downstream consumers outside this repo depend on the exact `sudo` command format (unlikely but possible if someone parses the prefix string).

### Feature Requests
- Auto-fetch the linked issue (from branch name pattern `fix/*-NNN`) to verify issue compliance
- Show `git log` for the branch to understand commit progression and verify the fix description matches the diff

## gemini [BLOCK]

### ERROR
**Verdict:** BLOCK

Model invocation failed: Model gemini failed: [31mGemini CLI is not running in a trusted directory. To proceed, either use `--skip-trust`, set the `GEMINI_CLI_TRUST_WORKSPACE=true` environment variable, or trust this directory in interactive mode. For more details, see https://geminicli.com/docs/cli/trusted-folders/#headless-and-automated-environments[0m

PR #143, opened 2026-05-28
Move gate cache to ~/.local/lib (no root required)

benthomasson:
## Summary

- Moves the cached gate .pyz from `/usr/local/lib/ftl2/gate.pyz` to `~/.local/lib/ftl2/gate.pyz`
- Non-root SSH users can now cache the gate without privilege escalation
- Resolves `/Users/ben` at runtime via shell expansion so the path works for any user

## Test plan

- [ ] Deploy to a server as non-root user, verify gate is cached at `~/.local/lib/ftl2/gate.pyz`
- [ ] Verify subsequent connections reuse the cached gate without re-uploading

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "ftl2"
-version = "0.3.3"
+version = "0.3.4"
 description = "Refactored faster-than-light automation framework with dataclasses and composition"
 authors = [{name = "Ben Thomasson", email = "ben.thomasson@gmail.com"}]
 readme = "README.md"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "ftl2"
-version = "0.3.4"
+version = "0.3.5"
 description = "Refactored faster-than-light automation framework with dataclasses and composition"
 authors = [{name = "Ben Thomasson", email = "ben.thomasson@gmail.com"}]
 readme = "README.md"
--- a/src/ftl2/ftl_modules/__init__.py
+++ b/src/ftl2/ftl_modules/__init__.py
@@ -20,7 +20,7 @@ Usage:
 from collections.abc import Callable
 from typing import Any
 
-from ftl2.ftl_modules.aws import ftl_ec2_instance
+from ftl2.ftl_modules.aws import ftl_ec2_instance, ftl_ec2_instance_info
 from ftl2.ftl_modules.command import ftl_command, ftl_shell
 from ftl2.ftl_modules.dnf import ftl_dnf
 from ftl2.ftl_modules.exceptions import (
@@ -59,6 +59,7 @@ FTL_MODULES: dict[str, ModuleFunc] = {
     "dnf5": ftl_dnf,
     "wait_for": ftl_wait_for,
     "ec2_instance": ftl_ec2_instance,
+    "ec2_instance_info": ftl_ec2_instance_info,
 }
 
 # Maps Ansible FQCNs to FTL module functions for compatibility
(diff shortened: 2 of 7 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#143
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [CONCERN]

### src/ftl2/runners.py:_update_gate_stable_path
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

**The core intent is good** — moving from `/usr/local/lib` (root-required) to `~/.local/lib` (user-writable) eliminates the need for root access during non-subsystem gate updates.

**Two correctness concerns:**

1. **Shell expansion via `echo $HOME`** (line 1345-1347): Using `echo $HOME` to expand the path works but is fragile. If `$HOME` is unset (e.g., some minimal containers, cron-over-SSH, or `become` scenarios where the env is stripped), this silently produces `.local/lib/ftl2/gate.pyz` as an absolute path rooted in whatever the cwd is. A safer approach would be `echo ~` or a Python-side `pathlib.PurePosixPath` with the home dir obtained via `getent passwd $(whoami)`. That said, `$HOME` is set in the vast majority of SSH sessions, so this is an edge case.

2. **`$(dirname ...)` inside the remote `conn.run`** (line 1350): This works but introduces a subshell expansion on the remote host. The previous code used Python's `os.path.dirname()` locally, which is arguably more predictable. The current approach is fine as long as the remote shell is POSIX-compliant, which it will be in practice.

**Integration gap — `_open_gate` and cleanup don't reference `GATE_USER_PATH`:**

- `_open_gate` (line 977) doesn't try the user path as a fallback when subsystem isn't available. The gate is run from the `~/.ftl` temp dir, so the stable path is only used for caching between connections — this is fine if that's the intent, but the user path is never actually *read from* in the current code, only written to. The purpose of the stable path update becomes unclear without a corresponding consumer.
- The `decommission_subsystem` method (line 1618) still only cleans up `GATE_SUBSYSTEM_PATH`, not `GATE_USER_PATH`. If a user runs decommission, the `~/.local/lib/ftl2/gate.pyz` will be left behind.

**No tests** — grep found no test coverage for `_update_gate_stable_path`. This is an SSH-heavy method so unit testing is harder, but at minimum a mock-based test verifying the correct remote commands are issued would catch regressions.

### src/ftl2/runners.py (class-level constants)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

`GATE_USER_PATH = ".local/lib/ftl2/gate.pyz"` is a relative path (no leading `~`), which is correct since it gets prefixed with `$HOME/` at the call site. `GATE_SUBSYSTEM_PATH` is preserved for the subsystem registration path, which still needs root — appropriate separation.

### Self-Review
**Limitations:** - Could not see whether `GATE_USER_PATH` is consumed anywhere else (e.g., in `_open_gate` or reconnect logic) — if nothing ever reads from this path, the upload is write-only caching that never gets used.
- No test files were found covering this method, so I cannot verify any test assertions.
- Could not inspect the git commit message for intent context beyond the diff.

### Feature Requests
- Show all callers and consumers of modified constants/methods to verify the feature is wired end-to-end (e.g., is `GATE_USER_PATH` ever read from, or only written to?)
- Include related test files even when they don't directly name the modified function — tests for the caller (`connect_or_reuse`) would be relevant here.

## gemini [BLOCK]

### ERROR
**Verdict:** BLOCK

Model invocation failed: Model gemini failed: [31mGemini CLI is not running in a trusted directory. To proceed, either use `--skip-trust`, set the `GEMINI_CLI_TRUST_WORKSPACE=true` environment variable, or trust this directory in interactive mode. For more details, see https://geminicli.com/docs/cli/trusted-folders/#headless-and-automated-environments[0m

PR #144, opened 2026-06-11
Add native async AWS modules (EC2 + Route 53)

benthomasson:
## Summary
- Implement `ftl_ec2_instance` and `ftl_ec2_instance_info` as native async modules using aioboto3
- Implement `ftl_route53_record` and `ftl_route53_info` for DNS record management
- Register all modules in the three registries (`aws/__init__.py`, `ftl_modules/__init__.py`, `executor.py`)
- Bump version 0.3.3 → 0.3.5

## Test plan
- [x] EC2: list instances, create, terminate against real AWS
- [x] Route 53: list zones, list records, create A record, delete A record against eemexpert.com
- [x] Full lifecycle validated with SAML authentication

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "ftl2"
-version = "0.3.3"
+version = "0.3.4"
 description = "Refactored faster-than-light automation framework with dataclasses and composition"
 authors = [{name = "Ben Thomasson", email = "ben.thomasson@gmail.com"}]
 readme = "README.md"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "ftl2"
-version = "0.3.4"
+version = "0.3.5"
 description = "Refactored faster-than-light automation framework with dataclasses and composition"
 authors = [{name = "Ben Thomasson", email = "ben.thomasson@gmail.com"}]
 readme = "README.md"
--- a/src/ftl2/ftl_modules/__init__.py
+++ b/src/ftl2/ftl_modules/__init__.py
@@ -20,7 +20,7 @@ Usage:
 from collections.abc import Callable
 from typing import Any
 
-from ftl2.ftl_modules.aws import ftl_ec2_instance
+from ftl2.ftl_modules.aws import ftl_ec2_instance, ftl_ec2_instance_info
 from ftl2.ftl_modules.command import ftl_command, ftl_shell
 from ftl2.ftl_modules.dnf import ftl_dnf
 from ftl2.ftl_modules.exceptions import (
@@ -59,6 +59,7 @@ FTL_MODULES: dict[str, ModuleFunc] = {
     "dnf5": ftl_dnf,
     "wait_for": ftl_wait_for,
     "ec2_instance": ftl_ec2_instance,
+    "ec2_instance_info": ftl_ec2_instance_info,
 }
 
 # Maps Ansible FQCNs to FTL module functions for compatibility
@@ -77,6 +78,7 @@ ANSIBLE_COMPAT: dict[str, ModuleFunc] = {
     "ansible.builtin.wait_for": ftl_wait_for,
     # AWS modules
     "amazon.aws.ec2_instance": ftl_ec2_instance,
+    "amazon.aws.ec2_instance_info": ftl_ec2_instance_info,
 }
 
 
@@ -166,4 +168,5 @@ __all__ = [
     "ftl_wait_for",
     # AWS modules
     "ftl_ec2_instance",
+    "ftl_ec2_instance_info",
 ]
--- a/src/ftl2/ftl_modules/__init__.py
+++ b/src/ftl2/ftl_modules/__init__.py
@@ -20,7 +20,12 @@ Usage:
 from collections.abc import Callable
 from typing import Any
 
-from ftl2.ftl_modules.aws import ftl_ec2_instance, ftl_ec2_instance_info
+from ftl2.ftl_modules.aws import (
+    ftl_ec2_instance,
+    ftl_ec2_instance_info,
+    ftl_route53_info,
+    ftl_route53_record,
+)
 from ftl2.ftl_modules.command import ftl_command, ftl_shell
 from ftl2.ftl_modules.dnf import ftl_dnf
 from ftl2.ftl_modules.exceptions import (
@@ -60,6 +65,8 @@ FTL_MODULES: dict[str, ModuleFunc] = {
     "wait_for": ftl_wait_for,
     "ec2_instance": ftl_ec2_instance,
     "ec2_instance_info": ftl_ec2_instance_info,
+    "route53_record": ftl_route53_record,
+    "route53_info": ftl_route53_info,
 }
 
 # Maps Ansible FQCNs to FTL module functions for compatibility
@@ -79,6 +86,8 @@ ANSIBLE_COMPAT: dict[str, ModuleFunc] = {
     # AWS modules
     "amazon.aws.ec2_instance": ftl_ec2_instance,
     "amazon.aws.ec2_instance_info": ftl_ec2_instance_info,
+    "amazon.aws.route53": ftl_route53_record,
+    "amazon.aws.route53_info": ftl_route53_info,
 }
 
 
@@ -169,4 +178,6 @@ __all__ = [
     # AWS modules
     "ftl_ec2_instance",
     "ftl_ec2_instance_info",
+    "ftl_route53_record",
+    "ftl_route53_info",
 ]
--- a/src/ftl2/ftl_modules/aws/__init__.py
+++ b/src/ftl2/ftl_modules/aws/__init__.py
@@ -4,5 +4,11 @@ Async AWS modules using aioboto3 for non-blocking cloud operations.
 """
 
 from ftl2.ftl_modules.aws.ec2 import ftl_ec2_instance, ftl_ec2_instance_info
+from ftl2.ftl_modules.aws.route53 import ftl_route53_info, ftl_route53_record
 
-__all__ = ["ftl_ec2_instance", "ftl_ec2_instance_info"]
+__all__ = [
+    "ftl_ec2_instance",
+    "ftl_ec2_instance_info",
+    "ftl_route53_record",
+    "ftl_route53_info",
+]
--- a/src/ftl2/ftl_modules/aws/__init__.py
+++ b/src/ftl2/ftl_modules/aws/__init__.py
@@ -3,6 +3,6 @@
 Async AWS modules using aioboto3 for non-blocking cloud operations.
 """
 
-from ftl2.ftl_modules.aws.ec2 import ftl_ec2_instance
+from ftl2.ftl_modules.aws.ec2 import ftl_ec2_instance, ftl_ec2_instance_info
 
-__all__ = ["ftl_ec2_instance"]
+__all__ = ["ftl_ec2_instance", "ftl_ec2_instance_info"]
--- a/src/ftl2/ftl_modules/aws/ec2.py
+++ b/src/ftl2/ftl_modules/aws/ec2.py
@@ -147,6 +147,21 @@ async def ftl_ec2_instance(
                             "state": current,
                             "instance": _extract_instance(existing),
                         }
+                    if current == "pending":
+                        if wait:
+                            await _wait_for_state(
+                                ec2, existing["InstanceId"], "running", wait_timeout,
+                            )
+                        resp = await ec2.describe_instances(
+                            InstanceIds=[existing["InstanceId"]],
+                        )
+                        inst = resp["Reservations"][0]["Instances"][0]
+                        return {
+                            "changed": False,
+                            "instance_id": inst["InstanceId"],
+                            "state": inst["State"]["Name"],
+                            "instance": _extract_instance(inst),
+                        }
                     if current in _STOPPED_STATES:
                         await ec2.start_instances(
                             InstanceIds=[existing["InstanceId"]],
@@ -197,6 +212,7 @@ async def ftl_ec2_instance(
                         ],
                     }]
 
+                run_params.update(kwargs)
                 resp = await ec2.run_instances(**run_params)
                 inst = resp["Instances"][0]
                 new_id = inst["InstanceId"]
--- a/src/ftl2/ftl_modules/aws/ec2.py
+++ b/src/ftl2/ftl_modules/aws/ec2.py
@@ -45,8 +45,11 @@ async def _find_instance(ec2, instance_id: str | None, name: str | None) -> dict
     if instance_id:
         try:
             resp = await ec2.describe_instances(InstanceIds=[instance_id])
-        except ClientError:
-            return None
+        except ClientError as e:
+            code = e.response["Error"]["Code"]
+            if code in ("InvalidInstanceID.NotFound", "InvalidInstanceID.Malformed"):
+                return None
+            raise
         for res in resp.get("Reservations", []):
             for inst in res.get("Instances", []):
                 if inst["State"]["Name"] not in _TERMINATED_STATES:
--- a/src/ftl2/ftl_modules/aws/ec2.py
+++ b/src/ftl2/ftl_modules/aws/ec2.py
@@ -29,7 +29,7 @@ def _extract_instance(instance: dict) -> dict[str, Any]:
         "subnet_id": instance.get("SubnetId"),
         "image_id": instance.get("ImageId"),
         "key_name": instance.get("KeyName"),
-        "launch_time": instance.get("LaunchTime", ""),
+        "launch_time": lt.isoformat() if (lt := instance.get("LaunchTime")) else "",
         "tags": tags,
         "security_groups": [
             {"id": sg["GroupId"], "name": sg["GroupName"]}
@@ -165,11 +165,9 @@ async def ftl_ec2_instance(
                             "instance": _extract_instance(inst),
                         }
                     if current == "stopping":
-                        if wait:
-                            await _wait_for_state(
-                                ec2, existing["InstanceId"], "stopped", wait_timeout,
-                            )
-                        # Fall through to start it below
+                        await _wait_for_state(
+                            ec2, existing["InstanceId"], "stopped", wait_timeout,
+                        )
                         current = "stopped"
                     if current in _STOPPED_STATES:
                         await ec2.start_instances(
--- a/src/ftl2/ftl_modules/aws/ec2.py
+++ b/src/ftl2/ftl_modules/aws/ec2.py
@@ -40,10 +40,12 @@ def _extract_instance(instance: dict) -> dict[str, Any]:
 
 async def _find_instance(ec2, instance_id: str | None, name: str | None) -> dict | None:
     """Find an existing instance by ID or Name tag, skipping terminated."""
+    from botocore.exceptions import ClientError
+
     if instance_id:
         try:
             resp = await ec2.describe_instances(InstanceIds=[instance_id])
-        except ec2.exceptions.ClientError:
+        except ClientError:
             return None
         for res in resp.get("Reservations", []):
             for inst in res.get("Instances", []):
@@ -162,6 +164,13 @@ async def ftl_ec2_instance(
                             "state": inst["State"]["Name"],
                             "instance": _extract_instance(inst),
                         }
+                    if current == "stopping":
+                        if wait:
+                            await _wait_for_state(
+                                ec2, existing["InstanceId"], "stopped", wait_timeout,
+                            )
+                        # Fall through to start it below
+                        current = "stopped"
                     if current in _STOPPED_STATES:
                         await ec2.start_instances(
                             InstanceIds=[existing["InstanceId"]],
@@ -180,6 +189,11 @@ async def ftl_ec2_instance(
                             "state": inst["State"]["Name"],
                             "instance": _extract_instance(inst),
                         }
+                    raise FTLModuleError(
+                        f"Instance {existing['InstanceId']} is in unexpected state: {current}",
+                        instance_id=existing["InstanceId"],
+                        current_state=current,
+                    )
 
                 if not image_id:
                     raise FTLModuleError(
--- a/src/ftl2/ftl_modules/aws/ec2.py
+++ b/src/ftl2/ftl_modules/aws/ec2.py
@@ -5,30 +5,350 @@ Async EC2 instance management using aioboto3.
 
 from typing import Any
 
-from ftl2.ftl_modules.exceptions import requires_extra
+from ftl2.ftl_modules.exceptions import FTLModuleError, requires_extra
 
-__all__ = ["ftl_ec2_instance"]
+__all__ = ["ftl_ec2_instance", "ftl_ec2_instance_info"]
+
+_RUNNING_STATES = ("running",)
+_STOPPED_STATES = ("stopped",)
+_TERMINATED_STATES = ("terminated", "shutting-down")
+_ACTIVE_STATES = ("pending", "running", "stopping", "stopped")
+
+
+def _extract_instance(instance: dict) -> dict[str, Any]:
+    """Normalize a boto3 instance dict to a flat result."""
+    tags_list = instance.get("Tags") or []
+    tags = {t["Key"]: t["Value"] for t in tags_list}
+    return {
+        "instance_id": instance["InstanceId"],
+        "instance_type": instance.get("InstanceType"),
+        "state": instance["State"]["Name"],
+        "public_ip": instance.get("PublicIpAddress"),
+        "private_ip": instance.get("PrivateIpAddress"),
+        "vpc_id": instance.get("VpcId"),
+        "subnet_id": instance.get("SubnetId"),
+        "image_id": instance.get("ImageId"),
+        "key_name": instance.get("KeyName"),
+        "launch_time": instance.get("LaunchTime", ""),
+        "tags": tags,
+        "security_groups": [
+            {"id": sg["GroupId"], "name": sg["GroupName"]}
+            for sg in instance.get("SecurityGroups", [])
+        ],
+    }
+
+
+async def _find_instance(ec2, instance_id: str | None, name: str | None) -> dict | None:
+    """Find an existing instance by ID or Name tag, skipping terminated."""
+    if instance_id:
+        try:
+            resp = await ec2.describe_instances(InstanceIds=[instance_id])
+        except ec2.exceptions.ClientError:
+            return None
+        for res in resp.get("Reservations", []):
+            for inst in res.get("Instances", []):
+                if inst["State"]["Name"] not in _TERMINATED_STATES:
+                    return inst
+        return None
+
+    if name:
+        resp = await ec2.describe_instances(Filters=[
+            {"Name": "tag:Name", "Values": [name]},
+            {"Name": "instance-state-name", "Values": list(_ACTIVE_STATES)},
+        ])
+        for res in resp.get("Reservations", []):
+            for inst in res.get("Instances", []):
+                return inst
+    return None
+
+
+async def _wait_for_state(ec2, instance_id: str, target: str, timeout: int) -> None:
+    """Wait for an instance to reach a target state."""
+    import asyncio
+
+    waited = 0
+    interval = 5
+    while waited < timeout:
+        resp = await ec2.describe_instances(InstanceIds=[instance_id])
+        state = resp["Reservations"][0]["Instances"][0]["State"]["Name"]
+        if state == target:
+            return
+        await asyncio.sleep(interval)
+        waited += interval
+    raise FTLModuleError(
+        f"Timed out waiting for instance {instance_id} to reach '{target}' "
+        f"after {timeout}s",
+        instance_id=instance_id,
+        target_state=target,
+    )
 
 
 @requires_extra("aws", "aioboto3")
 async def ftl_ec2_instance(
     instance_id: str | None = None,
+    name: str | None = None,
     state: str = "present",
     instance_type: str = "t3.micro",
     image_id: str | None = None,
+    key_name: str | None = None,
+    security_groups: list[str] | None = None,
+    vpc_subnet_id: str | None = None,
+    tags: dict[str, str] | None = None,
+    user_data: str | None = None,
+    wait: bool = True,
+    wait_timeout: int = 600,
+    region: str | None = None,
     **kwargs: Any,
 ) -> dict[str, Any]:
     """Manage EC2 instances.
 
     Args:
         instance_id: ID of existing instance
-        state: Desired state - present, absent, running, stopped
+        name: Name tag for idempotent instance lookup/creation
+        state: Desired state (present, running, started, stopped, terminated, absent)
         instance_type: EC2 instance type
         image_id: AMI ID for new instances
+        key_name: SSH key pair name
+        security_groups: List of security group IDs
+        vpc_subnet_id: VPC subnet ID
+        tags: Tags to apply (Name tag added automatically from name param)
+        user_data: User data script
+        wait: Wait for state transitions (default True)
+        wait_timeout: Timeout in seconds (default 600)
+        region: AWS region (defaults to AWS_DEFAULT_REGION or profile)
         **kwargs: Additional EC2 parameters
 
     Returns:
-        Result dict with instance_id and changed status
+        Result dict with changed, instance_id, state, and instance details
     """
-    # Placeholder - will be implemented when AWS native modules are built
-    raise NotImplementedError("ftl_ec2_instance is not yet implemented")
+    import aioboto3
+    from botocore.exceptions import BotoCoreError, ClientError
+
+    state = state.lower()
+    if state == "started":
+        state = "running"
+    if state == "absent":
+        state = "terminated"
+    if state == "present":
+        state = "running"
+
+    session = aioboto3.Session()
+    try:
+        async with session.client("ec2", region_name=region) as ec2:
+            existing = await _find_instance(ec2, instance_id, name)
+
+            if state in ("running",):
+                if existing:
+                    current = existing["State"]["Name"]
+                    if current in _RUNNING_STATES:
+                        return {
+                            "changed": False,
+                            "instance_id": existing["InstanceId"],
+                            "state": current,
+                            "instance": _extract_instance(existing),
+                        }
+                    if current in _STOPPED_STATES:
+                        await ec2.start_instances(
+                            InstanceIds=[existing["InstanceId"]],
+                        )
+                        if wait:
+                            await _wait_for_state(
+                                ec2, existing["InstanceId"], "running", wait_timeout,
+                            )
+                        resp = await ec2.describe_instances(
+                            InstanceIds=[existing["InstanceId"]],
+                        )
+                        inst = resp["Reservations"][0]["Instances"][0]
+                        return {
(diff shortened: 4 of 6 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#144
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [BLOCK]

### src/ftl2/ftl_modules/aws/ec2.py:_find_instance
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Line ~52 catches `ec2.exceptions.ClientError`, but `ClientError` is not a service-modeled exception — it lives in `botocore.exceptions`, not in the client's `.exceptions` namespace. This will raise `AttributeError` at runtime when a describe call fails for a non-existent instance ID. Should be `from botocore.exceptions import ClientError` (imported at top of the helper or in its body) and catch `ClientError` directly.

### src/ftl2/ftl_modules/aws/ec2.py:ftl_ec2_instance (running state handler)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

When the existing instance is in a transitional state not explicitly handled — e.g., `"stopping"` — none of the `if` branches match (`_RUNNING_STATES`, `"pending"`, `_STOPPED_STATES`), so execution falls through past the `if existing:` block to the **create new instance** path. This silently creates a duplicate instance instead of waiting for the transition to complete or raising an error. Should either handle transitional states or add an `else` that raises.

### src/ftl2/ftl_modules/aws/route53.py:ftl_route53_record (DELETE path)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The DELETE path reads `existing.get("ResourceRecords", [])` and passes it as `ResourceRecords` in the change batch. Alias records have no `ResourceRecords` — they have `AliasTarget` instead. Attempting to delete an alias record with an empty `ResourceRecords` list will fail. The code should detect alias records and include the `AliasTarget` key in the change set.

### src/ftl2/ftl_modules/aws/ec2.py:_extract_instance
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Clean normalization of boto3 instance dict. Minor note: `LaunchTime` is a `datetime` object that could cause JSON serialization issues downstream, but this follows existing module patterns and the framework may handle it.

### src/ftl2/ftl_modules/aws/ec2.py:ftl_ec2_instance_info
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Paginated `describe_instances` with filter translation is correct. Returns `changed: False` as expected for an info/query module. Filter value normalization (`v if isinstance(v, list) else [v]`) handles both single-string and list inputs properly.

### src/ftl2/ftl_modules/aws/route53.py:ftl_route53_record (UPSERT path)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Idempotency check compares sorted values and TTL before deciding to change — correct. FQDN normalization via `_normalize_record_name` is applied consistently. Zone resolution works for both zone name and direct hosted_zone_id. The `wait` path polling `get_change` for `INSYNC` is the standard Route 53 pattern.

### src/ftl2/ftl_modules/aws/route53.py:ftl_route53_info
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Pagination via `IsTruncated`/`NextRecordName`/`NextRecordType` is correct for the Route 53 API. Client-side filtering by record name and type works. Alias records are properly exposed in the output. Performance could be poor on very large zones with specific record filters (no early termination), but functionally correct.

### src/ftl2/ftl_modules/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

All three registries (`FTL_MODULES`, `ANSIBLE_COMPAT_MODULES`, `__all__`) are updated consistently. FQCN `amazon.aws.route53` maps to `ftl_route53_record`, matching the Ansible collection convention where `amazon.aws.route53` is the record management module.

### src/ftl2/ftl_modules/executor.py:_get_module
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Module registrations mirror `__init__.py` exactly — both short names and FQCNs match. The lazy-import pattern is maintained correctly.

### Self-Review
**Limitations:** Could not verify aioboto3's runtime behavior of `client.exceptions.ClientError` — assessment is based on boto3/botocore documentation and typical behavior. Could not run the test suite to confirm existing tests still pass with the rewritten EC2 module. Did not have access to the full exceptions.py to verify `FTLModuleMissingDependencyError` inheritance chain.

### Feature Requests
- Include related test files in the diff or observations when source files are modified, so reviewers can verify test assertions still match changed signatures/return types
- For new files, show whether a corresponding test file exists even if it wasn't modified (already done here via observations — good)
- Flag when `except SomeException` references an attribute that isn't imported or defined in scope

## gemini [CONCERN]

### pyproject.toml
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Version bump (0.3.3 -> 0.3.5) and the `aws` optional dependency correctly includes `aioboto3`.
---

### src/ftl2/ftl_modules/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The central module registry is correctly updated to expose the new AWS modules (`ec2_instance_info`, `route53_record`, `route53_info`). Both direct names and Ansible FQCN aliases (e.g., `amazon.aws.route53`) are properly mapped, ensuring compatibility.

### src/ftl2/ftl_modules/aws/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Correctly exports the new module functions for the `aws` subpackage, maintaining a clean public API for cloud operations.

### src/ftl2/ftl_modules/aws/ec2.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The implementation replaces the previous stub with functional async logic. Idempotency is well-handled: `_find_instance` supports both explicit `instance_id` and Name tag lookups, and state transitions (start/stop/terminate) include proper waiting logic. Normalization of the Boto3 response in `_extract_instance` is consistent. However, this non-trivial logic for cloud resource management is currently completely untested, which is a significant risk for core functionality.

### src/ftl2/ftl_modules/aws/route53.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

New implementation for DNS management. It correctly handles zone resolution (resolving names to IDs) and record idempotency by sorting values before comparison. The use of `NextRecordName/Type` for manual pagination in `ftl_route53_info` is robust. Like the EC2 module, this implementation lacks any unit or integration tests to verify its behavior against mocked or real AWS APIs.

### src/ftl2/ftl_modules/executor.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The `_get_module` helper is correctly updated to include the new modules and their FQCN mappings. This ensures the runtime execution engine can resolve these modules when they are used in automation tasks.

### Self-Review
**Limitations:** I could not verify the runtime behavior against a live AWS environment or a simulator like Moto. My assessment of "VALID" correctness is based on standard `aioboto3` patterns and logical flow analysis. I relied on the provided observations to confirm that `FTLModuleError` and `requires_extra` were integrated correctly with the existing framework.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#144
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/ftl_modules/aws/ec2.py:ftl_ec2_instance
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Two correctness issues:

1. **State normalization uses sequential `if` instead of `elif` (lines 129-135)**. `state="present"` becomes `"running"` after line 134, which is correct, but `state="absent"` becomes `"terminated"` and then doesn't hit line 134, which is also fine. However, `state="started"` → `"running"` → then line 132 `if state == "absent"` is false → line 134 `if state == "present"` is false. So this works by accident, but an `elif` chain would be clearer and safer against future additions.

2. **`_wait_for_state` checks state before the first sleep (line 73-77)** — this is correct. However, the timeout is measured by wall-clock intervals and doesn't account for API latency. For a 600s default that's acceptable, but worth noting.

3. **Transitional state handling for `stopped` target (line 198-213)**: When an instance is in "pending" or "shutting-down" state and you request `state=stopped`, the code calls `stop_instances` directly without waiting for a stable state first. AWS will likely reject this with a ClientError, which gets caught and re-raised as `FTLModuleError`. Not incorrect, but fragile compared to the `running` state handler which handles transitional states explicitly.

4. **`launch_time` serialization (line 32)**: `instance.get("LaunchTime", "")` — `LaunchTime` is a `datetime` object from boto3, not a string. This will put a raw `datetime` in the return dict, which may fail JSON serialization downstream (e.g., audit recording, state files). Should call `.isoformat()` on it.

### src/ftl2/ftl_modules/aws/ec2.py:ftl_ec2_instance_info
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Clean read-only info module. Uses pagination correctly. Filter conversion from `dict[str, str|list[str]]` to boto3 Filter format is correct. Same `launch_time` serialization concern from `_extract_instance` applies here too.

### src/ftl2/ftl_modules/aws/route53.py:ftl_route53_record
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Two issues:

1. **`type` parameter shadows the builtin (line 46)**. The parameter is named `type` which shadows Python's `type()` builtin. While it doesn't cause a bug in this function's body, it's a code smell. Consider `record_type` as the parameter name to match Ansible's conventions better.

2. **DELETE uses stale TTL and records from lookup (lines 109-113)**: When deleting, the code fetches the existing record and uses its TTL and ResourceRecords for the DELETE ChangeBatch. This is correct — Route 53 requires the exact record set for DELETE operations. However, there's a TOCTOU race if the record changes between the `_find_record` lookup and the `change_resource_record_sets` call. This is inherent to the API and acceptable, but the error message from AWS in that case would be opaque.

3. **Alias record handling gap**: `_find_record` returns the full record set dict. If the existing record is an alias record (no `ResourceRecords`, has `AliasTarget`), the DELETE path at line 112 gets `resource_records = []` and sends an empty `ResourceRecords` list, which AWS will reject. The record module doesn't support creating alias records, but it could encounter them on DELETE.

### src/ftl2/ftl_modules/aws/route53.py:ftl_route53_info
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Correct pagination over `list_resource_record_sets`. Handles alias records in output (line 259-264). Name/type filtering after fetch is correct since `StartRecordName` is a starting point, not an exact filter. Stripping trailing dot from names (line 249) is a good UX choice for consistency with input.

### src/ftl2/ftl_modules/aws/route53.py:_resolve_zone_id
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Correctly normalizes zone name with trailing dot, uses `list_hosted_zones_by_name` with `MaxItems="1"`, and verifies exact match. Extracts zone ID from the `/hostedzone/XXXX` format correctly.

### src/ftl2/ftl_modules/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

All three registration points are updated correctly: `FTL_MODULES` (short names), `ANSIBLE_COMPAT` (FQCNs), and `__all__`. The FQCN `amazon.aws.route53` maps to `ftl_route53_record`, which matches the Ansible collection naming where `amazon.aws.route53` is the record management module. Existing tests cover `ftl_ec2_instance` FQCN lookup but not the new modules.

### src/ftl2/ftl_modules/executor.py:_get_module
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Module registry in `_get_module` correctly mirrors the `__init__.py` registrations. Both short names and FQCNs are present. The duplicate registry (here vs `__init__.py`) exists to break circular imports, which is documented.

### Self-Review
**Limitations:** Could not run the test suite to verify nothing is broken by the imports. Could not verify whether the audit recording path actually calls `json.dumps` on the result dict (which would expose the `datetime` serialization issue). Did not have access to the Ansible `amazon.aws.route53` module docs to verify FQCN mapping accuracy.

### Feature Requests
- Include the existing test file contents (e.g., `tests/test_ftl_modules.py`) in the diff or observations so I can verify whether existing test assertions still pass with the new imports
- Show the audit recording codepath to verify return dict serialization requirements

## gemini [CONCERN]

### pyproject.toml
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: CONSISTENT
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Standard version bump from 0.3.3 to 0.3.5 to reflect the addition of new AWS native modules.
---

### src/ftl2/ftl_modules/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly registers the new AWS modules (`ec2_instance_info`, `route53_record`, `route53_info`) in the `FTL_MODULES` registry and maps them to their respective Ansible FQCNs in `ANSIBLE_COMPAT`. Existing tests in `tests/test_ftl_modules.py` verify that these mappings are correctly established.

### src/ftl2/ftl_modules/aws/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Correctly exports the new module functions from the `aws` subpackage.

### src/ftl2/ftl_modules/aws/ec2.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

1. **Logic Issue**: In `ftl_ec2_instance`, the transition from `stopping` to `running` (lines 142-148) incorrectly sets `current = "stopped"` even if `wait=False`. This will cause a `ClientError` (IncorrectInstanceState) when `ec2.start_instances` is called on line 150, as the instance is still `stopping`.
2. **Untested**: While the implementation logic for idempotency and state transitions is comprehensive, there are no functional tests provided for these new features. Only the "missing dependency" behavior was previously covered by stubs.
3. **Serialization**: `_extract_instance` (line 34) returns `LaunchTime` as a `datetime` object. If this result is ever serialized to JSON (e.g., for reporting or cross-process communication), it will fail without a custom encoder.

### src/ftl2/ftl_modules/aws/route53.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

1. **Ambiguity**: `_resolve_zone_id` (line 17) uses `MaxItems="1"` and picks the first result. AWS allows multiple hosted zones with the same name; this implementation may non-deterministically pick the wrong zone if duplicates exist.
2. **Untested**: Like the EC2 module, this new file lacks functional tests to verify DNS record management logic, idempotency, or propagation waiting.
3. **Correctness**: The `DELETE` logic correctly fetches the existing record's values to satisfy Route 53's strict deletion requirements.

### src/ftl2/ftl_modules/executor.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly updates the internal `_get_module` registry. This is essential for the framework's runtime execution and avoids circular imports by localizing module registration.

### Self-Review
**Limitations:** - Functional verification of AWS modules was not possible without a live environment or a mock-based test suite (e.g., using `moto` or `aioboto3` mocks), which were not present in the changes.
- Review focused on logic flow and architectural integration within the FTL framework.

### Feature Requests
- Include mock-based unit tests for new cloud modules (EC2, Route53) to verify idempotency and error handling logic without requiring live AWS credentials.
- Add a utility to handle datetime serialization in module results to ensure JSON compatibility.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#144
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## claude [BLOCK]

### src/ftl2/ftl_modules/aws/ec2.py:_find_instance
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Catches **all** `ClientError` at line ~52, returning `None` for any failure — including `AuthFailure`, `UnauthorizedAccess`, and `RequestLimitExceeded`. Only `InvalidInstanceID.NotFound` / `InvalidInstanceID.Malformed` should be swallowed. A permissions error would silently create a duplicate instance instead of surfacing the auth problem. Fix: check `e.response['Error']['Code']` before returning None.

### src/ftl2/ftl_modules/aws/ec2.py:ftl_ec2_instance (stopped state)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The `stopped` branch (around line 210) doesn't handle transitional states. `_find_instance` can return instances in "stopping" or "pending" state (per `_ACTIVE_STATES`). Calling `stop_instances` on a "stopping" instance succeeds but incorrectly reports `changed: True`. Should check for "stopping" (wait then return `changed: False`) and "pending" (arguably should wait for "running" first, then stop) the same way the "running" branch handles "stopping" and "pending".

### src/ftl2/ftl_modules/aws/ec2.py:ftl_ec2_instance (terminated state)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

When `existing` is None (line ~234), the return dict sets `"instance_id": instance_id` which could be `None`. Other code paths always return a string `instance_id`. Additionally, `_extract_instance(existing)` is called on the pre-termination snapshot rather than re-fetching — the returned `state` field in the instance dict will show the old state, not "terminated" or "shutting-down", which is inconsistent with the top-level `"state": "terminated"` in the same return dict.

### src/ftl2/ftl_modules/aws/ec2.py:ftl_ec2_instance (overall structure)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

State normalization (started→running, absent→terminated, present→running) is correct. The running branch properly handles pending, stopping→stopped→start, and stopped→start transitions. Instance creation with TagSpecifications, kwargs pass-through, and wait logic are all sound. The `_extract_instance` helper properly normalizes boto3 responses. The broad exception handler at the bottom re-raises `FTLModuleError` before catching other exceptions, which is correct.

### src/ftl2/ftl_modules/aws/ec2.py:ftl_ec2_instance_info
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Clean read-only info module. Paginator usage handles large result sets correctly. Filter transformation from `dict[str, str | list[str]]` to AWS filter format is correct. Always returns `changed: False`.

### src/ftl2/ftl_modules/aws/route53.py:ftl_route53_record
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

UPSERT/DELETE logic is correct. Idempotency check compares sorted values + TTL before deciding to skip. DELETE correctly fetches existing ResourceRecords and TTL (required by AWS API for exact-match delete). Zone resolution with trailing-dot normalization is handled. Wait-for-propagation loop is straightforward. The `type` parameter shadows the builtin but matches Ansible convention — acceptable.

### src/ftl2/ftl_modules/aws/route53.py:ftl_route53_info
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

When `type` is provided without `record`, `StartRecordType` is set in the API params (line ~260) but per AWS docs, `StartRecordType` is ignored unless `StartRecordName` is also present. The client-side filter at line ~276 still works correctly, but the code gives the false impression that the API is doing server-side filtering. More practically: pagination continues fetching the entire zone when only one record is needed — once the post-fetch filter stops matching, it should `break` rather than continue paging.

### src/ftl2/ftl_modules/aws/route53.py:_resolve_zone_id
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Correctly normalizes zone name with trailing dot, uses `list_hosted_zones_by_name` with `MaxItems="1"`, and validates the returned zone name matches exactly (since the API returns zones starting from the given name alphabetically). Zone ID extraction via `split("/")[-1]` correctly strips the `/hostedzone/` prefix.

### src/ftl2/ftl_modules/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

All three registries (`FTL_MODULES`, `ANSIBLE_COMPAT_MODULES`, `__all__`) are updated consistently. FQCN mapping `amazon.aws.route53` → `ftl_route53_record` matches Ansible's convention where the route53 module manages records while route53_info queries them.

### src/ftl2/ftl_modules/executor.py:_get_module
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Module map in `_get_module` matches the registries in `__init__.py`. Both short names and FQCNs are present. The duplicate registry pattern (here + `__init__.py`) exists to avoid circular imports and is a pre-existing pattern — not introduced by this PR.

### Test coverage (overall)
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Zero tests for any of the new modules. The observation data confirms `ec2_tests.test_count: 0` and `route53_tests.test_count: 0`. The existing test file only checks `ftl_ec2_instance` is callable and that it raises when aioboto3 is missing. At minimum, the new modules need: (1) callable existence tests like the existing `test_ftl_ec2_instance_exists`, (2) missing-dependency tests via `requires_extra`, (3) mocked aioboto3 tests for the core state machine logic (especially the EC2 state transitions which have significant branching). The EC2 module alone has ~200 lines of untested state machine logic.

### Self-Review
**Limitations:** Could not verify aioboto3 paginator async iteration behavior — assumed it works like documented. Could not see the full `tests/test_ftl_modules.py` to verify whether the existing test assertions would need updates for the expanded EC2 implementation (the stub previously raised `NotImplementedError`, and the test catches `(NotImplementedError, FTLModuleError)` — this still works but the `NotImplementedError` branch is now dead). Did not have access to verify whether `wait_for` or other existing module patterns use similar polling strategies for comparison.

### Feature Requests
- Include related test files in full when source files have `test_count > 0` in observations — I could see test callers but not the full test structure to verify nothing breaks
- Show AWS SDK method signatures or link to docs for `describe_instances`, `list_hosted_zones_by_name` etc. to verify parameter correctness without relying on reviewer's memory

## gemini [CONCERN]

### src/ftl2/ftl_modules/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly registers the new AWS modules in both the internal `FTL_MODULES` registry and the `ANSIBLE_COMPAT` mapping.

### src/ftl2/ftl_modules/aws/ec2.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The implementation of `ftl_ec2_instance` provides robust state transition logic (handling intermediate states like `stopping` or `pending`) and idempotent lookup by `Name` tag. However, this logic is entirely untested in the provided changes. Additionally, the `user_data` parameter is passed directly to `run_instances`; in many AWS SDK versions, this requires manual base64 encoding by the caller or the module to avoid API errors or corrupted data on the instance.

### src/ftl2/ftl_modules/aws/route53.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Correctly implements DNS record management with idempotency checks for values and TTL. It uses appropriate Route53 API calls (`list_resource_record_sets`, `change_resource_record_sets`) and handles pagination. Similar to the EC2 module, there is a total lack of unit tests for the new logic. Furthermore, `ftl_route53_record` currently only supports standard `ResourceRecords` and lacks support for `AliasTarget` records, which is a significant limitation for AWS DNS management.

### src/ftl2/ftl_modules/executor.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The internal module registry `_get_module` was correctly updated to facilitate runtime loading of the new AWS modules, avoiding circular imports.

### Self-Review
**Limitations:** I was unable to verify the functional correctness of the AWS state machines (especially state transitions like `stopping` -> `stopped` -> `running`) without mocks or a live environment. I also could not verify if `aioboto3` handles the base64 encoding of `UserData` automatically in this specific context, as behavior can vary between SDK versions and specific API operations.

### Feature Requests
- Include mock-based unit tests for new cloud modules to verify state machine logic and idempotency without requiring live credentials.
- Add support for Route53 Alias records in `ftl_route53_record`.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#144
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/ftl_modules/aws/ec2.py:ftl_ec2_instance
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

1. **State mutation before validation** (ec2.py ~line 131-136): `state` is normalized via three sequential `if` statements (`started→running`, `absent→terminated`, `present→running`). These are `if` not `elif`, so all three execute. Functionally this works because the transformations don't chain, but it reads as fragile — a future addition could accidentally chain. An `elif` or match/case would be clearer.

2. **`_wait_for_state` polling is wall-clock-imprecise**: The `waited += interval` approach doesn't account for the time spent in `describe_instances`. Under high latency, the effective timeout can be significantly longer than `wait_timeout`. Minor, but worth noting for a 600s default.

3. **`_find_instance` returns the first instance when searching by Name tag** (ec2.py ~line 62-66): If multiple non-terminated instances share the same Name tag, the function silently picks whichever AWS returns first. This is the common Ansible behavior, but it's a possible surprise for users with duplicate names.

4. **Stopping-state handling is good** (ec2.py ~line 158-163): The code waits for `stopping→stopped` before issuing `start_instances` — this avoids the `IncorrectInstanceState` error from AWS. Well done.

5. **Catch-all `except Exception`** (ec2.py ~line 315-320): Wrapping all exceptions into `FTLModuleError` loses stack trace fidelity. The `from e` preserves the chain, but the generic message `"EC2 operation failed"` obscures the root cause for non-AWS exceptions (e.g., `TypeError` from bad arguments). Consider letting non-AWS exceptions propagate naturally.

### src/ftl2/ftl_modules/aws/ec2.py:ftl_ec2_instance_info
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Clean info-gathering module. Uses paginator correctly. Filter transformation from `dict[str, str | list[str]]` to the AWS `Filters` format is correct. Same catch-all `except Exception` concern as above, but lower risk for a read-only operation.

### src/ftl2/ftl_modules/aws/route53.py:ftl_route53_record
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

1. **Parameter name shadows builtin** (route53.py line 38): `type: str = "A"` shadows the built-in `type()`. This is a minor Python anti-pattern. Consider `record_type` as the parameter name (the Ansible module uses `type`, so there's a compatibility argument, but it's still a concern with ruff/pylint).

2. **DELETE uses stale TTL/records from lookup** (route53.py ~line 112-117): When `state=absent`, the code fetches the existing record and uses its `ResourceRecords` and `TTL` for the DELETE ChangeBatch. This is correct — Route 53 DELETE requires an exact match of the record set. Good.

3. **Idempotency check compares sorted values but not record type aliases** (route53.py ~line 100-106): The idempotency check compares `sorted(values) == sorted(existing_values)` and `existing.get("TTL") == ttl`. This is correct for simple records. However, for alias records (`AliasTarget`), the code would try to upsert a `ResourceRecords`-based record over an alias, which AWS would reject. There's no guard against this case.

4. **`_resolve_zone_id` trusts `list_hosted_zones_by_name` ordering** (route53.py ~line 13-21): The function requests `MaxItems="1"` and checks exact name match. This is correct — `list_hosted_zones_by_name` returns zones in lexicographic order starting from the given name, so an exact match will always be first if it exists.

5. **`_find_record` doesn't handle alias records** (route53.py ~line 170-179): If the existing record is an alias, it has no `ResourceRecords` key, which means the idempotency check in the main function would compare `sorted([])` vs `sorted(values)`, always seeing a change and upserting, which would replace the alias with a standard record. This is a correctness concern for existing alias records.

### src/ftl2/ftl_modules/aws/route53.py:ftl_route53_info
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The pagination loop is correct (checks `IsTruncated`, uses `NextRecordName`/`NextRecordType`). Alias records are included with their target info. The filtering by `record` and `type` is done client-side after fetching, which is fine given `StartRecordName` is a cursor, not a filter. The `.rstrip(".")` normalization on `entry["name"]` allows callers to match against non-FQDN names.

### src/ftl2/ftl_modules/aws/ec2.py:_extract_instance
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Handles `LaunchTime` serialization correctly with the walrus operator and `.isoformat()`. Tags normalization from list-of-dicts to flat dict is correct. Security groups extraction is clean.

### src/ftl2/ftl_modules/__init__.py (registry updates)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

All three new modules are registered in `FTL_MODULES` (short names), `ANSIBLE_FQCN_MODULES` (FQCN mapping), and the `__all__` export list. The FQCN `"amazon.aws.route53"` mapping to `ftl_route53_record` matches the Ansible collection's FQCN. The `list_ansible_compat` function is also updated. Existing tests import `ftl_ec2_instance` and pass, but no tests cover the three new entries.

### src/ftl2/ftl_modules/executor.py:_get_module
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Mirrors the `__init__.py` registry exactly. The duplication between `__init__.py` and `executor.py` exists to avoid circular imports — this is an existing pattern, not introduced by this PR. The existing `test_get_module_by_fqcn` test covers the `amazon.aws.ec2_instance` path but not the three new entries.

### Self-Review
**Limitations:** Could not run the test suite to verify existing tests still pass with the new imports. Could not verify whether `aioboto3`'s async paginator actually supports the `async for` pattern used in `ftl_ec2_instance_info` (it does in recent versions, but the minimum version pin is `>=11.0.0` which should be fine). Did not have access to the full `ftl_ec2_instance` function in the original code to compare the scope of the rewrite (the original was a stub `raise NotImplementedError`).

### Feature Requests
- Include the existing test file in full when new functions are added to source files that already have test coverage, so reviewers can verify the tests still compile with changed imports
- Show the `git log --oneline` for the branch to understand the version skip (0.3.3 → 0.3.5)

## gemini [CONCERN]

### pyproject.toml
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: MEETS
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: CONSISTENT
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Version bump to 0.3.5 is consistent with adding new features. Dependency management for AWS (`aioboto3`) is correctly placed in optional-dependencies.
---

### src/ftl2/ftl_modules/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly imports and registers the new AWS modules in both the short-name and FQCN registries. The `list_ansible_compat` function is also properly updated.

### src/ftl2/ftl_modules/aws/ec2.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** UNTESTED
**Integration:** WIRED

The implementation is logically sound and follows async best practices. It correctly handles state transitions (e.g., waiting for 'stopped' before 'starting' if an instance is 'stopping'). However, there are no unit tests provided that mock `aioboto3` to verify these state transitions and idempotency logic.

### src/ftl2/ftl_modules/aws/route53.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** UNTESTED
**Integration:** WIRED

The logic for DNS record management is correct, including robust idempotency checks and FQDN normalization. It correctly handles Route 53's requirement for exact matches on DELETE actions. Like the EC2 module, it lacks unit tests for the actual AWS interaction logic. Note: The record module currently only supports standard ResourceRecords and does not yet handle AliasTarget records.

### src/ftl2/ftl_modules/executor.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** MEETS
**Test Coverage:** COVERED
**Integration:** WIRED

The local module registry used for execution is correctly updated with the new AWS modules, including FQCN mappings for Ansible compatibility. This is crucial for the framework's ability to dispatch these modules at runtime.

### Self-Review
**Limitations:** I was unable to verify the exact definition of `FTLModuleError` in `src/ftl2/ftl_modules/exceptions.py` because the observation tool failed to find it as a function (likely because it is a class). I assumed it supports the keyword arguments passed in the implementation based on standard project patterns. I also could not verify if `aioboto3` was actually available in the environment to run the existing stub tests.

### Feature Requests
- Improve `symbol_search` or `read_file` to distinguish between classes and functions when the user provides a name without context.
- Automatically provide the full body of the `exceptions.py` or similar base files when reviewing modules that depend heavily on project-specific error types.

PR #147, opened 2026-06-11
Fix copy ownership under become: default to become_user

### LINKED ISSUE #145: copy module writes files with connecting user ownership instead of root under become=True
## Bug

When using become=True with a non-root user (e.g., fedora), the copy module writes files owned by the connecting user instead of root. This causes problems for files that require root ownership, like /etc/sudoers.d/ entries.

## Reproduction

```python
ftl.add_host(hostname=HOST, ansible_host=ip, ansible_user='fedora', ansible_become=True)

await ftl[HOST].copy(
    content='admin ALL=(ALL) NOPASSWD: ALL\n',
    dest='/etc/sudoers.d/admin',
    mode='0440',
)
```

Expected: file owned by root:root
Actual: file owned by fedora:fedora (uid 1000)

This breaks sudo because sudoers files must be owned by uid 0:
```
sudo: /etc/sudoers.d/admin is owned by uid 1000, should be 0
```

## Context

Discovered while deploying to AWS EC2 with Fedora Cloud AMI. The default fedora user has passwordless sudo via cloud-init, so become=True should escalate to root for all file operations including ownership.

The Linode deployment does not hit this because it connects as root directly.

## Workaround

Explicitly set owner and group on copy calls for system files:
```python
await ftl[HOST].copy(
    content='admin ALL=(ALL) NOPASSWD: ALL\n',
    dest='/etc/sudoers.d/admin',
    owner='root', group='root',
    mode='0440',
)
```

benthomasson:
## Summary

- When `become=True` and no explicit `owner`/`group` is specified, the copy method's SFTP-then-sudo-mv path left files owned by the connecting user instead of the become_user
- Now defaults owner/group to `become_cfg.become_user` (typically `root`) so files get correct ownership automatically
- Matches Ansible behavior: files created under become are owned by the become user unless overridden

## Root cause

The become path in `proxy.py`'s `_copy_to_host` writes the file via SFTP (as the connecting user), then `sudo mv`s it to the destination. `mv` preserves original ownership, so the file stays owned by the connecting user. The old code only ran `chown` when owner/group were explicitly provided — when omitted, no ownership fix happened.

## Test plan

- [x] All 2034 existing tests pass (1 pre-existing unrelated failure in test_gate_timeouts excluded)
- [ ] Integration test: deploy with `become=True` as non-root user, verify system files are owned by root

Fixes #145

### DIFF
--- a/src/ftl2/automation/proxy.py
+++ b/src/ftl2/automation/proxy.py
@@ -460,13 +460,12 @@ class HostScopedProxy:
                     if mode:
                         await ssh.run(become_cfg.become_prefix(f"chmod {mode} {quoted_dest}"))
 
-                    # Set ownership via sudo
-                    if owner and group:
-                        await ssh.run(become_cfg.become_prefix(f"chown {owner}:{group} {quoted_dest}"))
-                    elif owner:
-                        await ssh.run(become_cfg.become_prefix(f"chown {owner} {quoted_dest}"))
-                    elif group:
-                        await ssh.run(become_cfg.become_prefix(f"chgrp {group} {quoted_dest}"))
+                    # Set ownership via sudo — default to become_user when not specified
+                    effective_owner = owner or become_cfg.become_user
+                    effective_group = group or become_cfg.become_user
+                    await ssh.run(become_cfg.become_prefix(
+                        f"chown {effective_owner}:{effective_group} {quoted_dest}"
+                    ))
                 else:
                     # Direct SFTP path (original behavior, no sudo needed)
 


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#147
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/automation/proxy.py:_copy_to_host
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

**Correctness — mostly right, but one behavioral concern:**

The change correctly defaults file ownership to `become_cfg.become_user` when `owner`/`group` aren't specified. This aligns with Ansible semantics: when become is active, the file gets `mv`'d by sudo, making it root-owned; defaulting ownership to `become_user` is the right fix.

However, the old code was **opt-in**: omitting `owner` and `group` meant "don't touch ownership." The new code is **always-on**: `chown` runs unconditionally on every copy under become, even when the user deliberately omitted ownership params to mean "leave it alone." This is a semantic change, not just a default change. If someone runs:

```python
await ftl.webservers.copy(content="x", dest="/etc/foo", become=True)
```

Before: no chown. After: `chown root:root` (since `become_user` defaults to `"root"`).

When `become_user="root"` (the default), this is harmless since sudo already made the file root-owned. But the intent shift matters: there's no way to say "use become but don't set ownership."

**The `chown` also runs when `changed=False`** (idempotent no-op). This was true for the old ownership block too, so it's not a regression — but it means every become-path copy pays an extra SSH round-trip even when nothing changed, which wasn't the case before when `owner`/`group` were omitted.

**No test coverage for the new behavior.** All existing `test_copy_multi_host.py` tests use `become=False` mocks (line 15: `_make_host_config(name, become=False)`). There are zero tests that exercise the become path of `_copy_to_host` with or without explicit `owner`/`group`, so the new default-ownership logic is completely untested. I'd expect at least:
1. Copy under become without owner/group → chown uses become_user
2. Copy under become with explicit owner/group → chown uses those values
3. Copy under become with become_user="deploy" → chown uses deploy:deploy

### Self-Review
**Limitations:** Could not verify whether any integration/E2E tests exercise the become+copy path end-to-end. The observation for test file existence returned `exists: false` for all test files (appears to be a bug in the observation — those files do exist), so I had to discover them manually. Could not see the git history for what issue motivated this change beyond the commit message.

### Feature Requests
- Include the git commit message and any linked issue body as context for reviews, since the motivation behind a change is critical for assessing correctness
- When `copy_test_coverage` observation reports test files, verify the paths are correct (the observation returned mangled paths without dots)

## gemini [CONCERN]

### src/ftl2/automation/proxy.py:_copy_to_host
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The change forces ownership to the `become_user` when using `become`, even if `owner` or `group` were not specified. This introduces several regressions and questionable patterns:
1. **Idempotency Violation**: The `chown` command is now run unconditionally on every execution when `become` is active. This is inefficient and inconsistent with the non-become path (lines 497-512), which performs a `stat` check to avoid redundant changes.
2. **Missing Change Tracking**: The `changed` flag is not updated to `True` when ownership is modified. If the file content is identical but ownership is wrong, the task will return `changed: false` despite modifying the system state.
3. **Risky Group Defaulting**: Defaulting the group to the user's name (`effective_group = group or become_cfg.become_user`) assumes a group exists with that name. While common, this is not a universal guarantee across all Unix-like systems. A safer default for "user's primary group" is often achieved by leaving the group empty in `chown` (e.g., `chown user:`) or explicitly querying it.
4. **Behavioral Regression**: The previous implementation only touched ownership if explicitly requested via the `owner` or `group` parameters. This change forces ownership even when the user might have intended to preserve the "natural" ownership resulting from the `mv` command.

### Self-Review
**Limitations:** I do not have the full definition of the `BecomeConfig` class to confirm what `become_user` defaults to when not explicitly provided in the inventory/task, though I've assumed it defaults to "root" based on common patterns in the file.

### Feature Requests
- Provide the full source of related test files when possible (e.g., `tests/test_become.py`) to verify if the new logic breaks existing assertions about idempotency or result dictionaries.
- Include the definitions of referenced internal classes (like `BecomeConfig`) to better understand the state of variables like `become_user`.

PR #148, opened 2026-06-11
Fix gate setup under become: thread become config to _connect_gate

### LINKED ISSUE #146: service module gate setup fails under become=True with non-root user
## Bug

The `service` module fails when connecting as a non-root user with `ansible_become=True`. The gate setup command `mkdir -p ~/.ftl && chmod 700 ~/.ftl` returns exit code 1.

## Reproduction

```python
ftl.add_host(hostname=HOST, ansible_host=ip, ansible_user='admin', ansible_become=True)

# This fails:
await ftl[HOST].service(name='expert-service', state='restarted', enabled=True)
```

Error:
```
Failed to update gate at stable path: Process exited with non-zero exit status 1
asyncssh.process.ProcessError: Process exited with non-zero exit status 1
```

The gate setup runs `mkdir -p ~/.ftl && chmod 700 ~/.ftl` which fails because:
- The `~` resolves to the connecting user's home directory, not root's
- Under SELinux enforcing, the home directory may have restrictive labels
- If the home directory ownership is wrong (see #145), mkdir fails with Permission denied

## Impact

All `service` module calls fail under become. Workaround is to use `shell` with systemctl directly:

```python
# Instead of:
await ftl[HOST].service(name='expert-service', state='restarted', enabled=True)

# Use:
await ftl[HOST].shell(cmd='systemctl enable --now expert-service && systemctl restart expert-service')
```

## Context

Discovered while deploying expert-service to AWS EC2 with Fedora Cloud AMI. The `shell` and `copy` modules work fine under the same become configuration. Related to #145 (copy module ownership under become).

benthomasson:
## Summary

- The `become` config was extracted in `run()` but never passed through `_get_or_create_gate()` to `_connect_gate()`, so the gate process launched without sudo — all gate-based modules (service, etc.) failed under `become=True` with non-root users
- Threads `become` through the full call chain: `run()` → `_get_or_create_gate()` → `_connect_gate()` → `_open_gate()`
- Wraps `_update_gate_stable_path` commands with `become_prefix` since `/usr/local/lib/ftl2/` requires root
- Uses SFTP-to-temp then sudo-mv pattern for stable path updates under become (SFTP runs as connecting user)

## Test plan

- [ ] Integration test: redeploy to AWS using `service` module calls instead of `shell` systemctl workarounds

Fixes #146

### DIFF
--- a/src/ftl2/runners.py
+++ b/src/ftl2/runners.py
@@ -1358,7 +1358,8 @@ class RemoteModuleRunner(ModuleRunner):
             await conn.run(_cmd(f"mkdir -p {os.path.dirname(dest)}"), check=True)
             if use_become:
                 # SFTP runs as connecting user — upload to temp, then sudo mv
-                sftp_tmp = f"/tmp/ftl2-gate-{os.getpid()}.pyz.tmp"
+                import uuid
+                sftp_tmp = f"/tmp/ftl2-gate-{uuid.uuid4().hex}.pyz.tmp"
                 async with conn.start_sftp_client() as sftp:
                     await sftp.put(gate_local_path, sftp_tmp)
                 await conn.run(_cmd(f"chmod 755 {sftp_tmp}"), check=True)
--- a/src/ftl2/runners.py
+++ b/src/ftl2/runners.py
@@ -635,7 +635,7 @@ class RemoteModuleRunner(ModuleRunner):
 
             # Get or create gate connection
             gate = await self._get_or_create_gate(
-                cache_key, ssh_host, ssh_port, ssh_user, ssh_password, ssh_key_file, interpreter, context, host_name=host.name
+                cache_key, ssh_host, ssh_port, ssh_user, ssh_password, ssh_key_file, interpreter, context, host_name=host.name, become=become
             )
 
             try:
@@ -689,6 +689,7 @@ class RemoteModuleRunner(ModuleRunner):
         interpreter: str,
         context: ExecutionContext,
         host_name: str = "localhost",
+        become: "BecomeConfig | None" = None,
     ) -> Gate:
         """Get cached gate or create new one.
 
@@ -702,6 +703,7 @@ class RemoteModuleRunner(ModuleRunner):
             interpreter: Remote Python interpreter path
             context: Execution context with gate config
             host_name: Logical host name for gate-side policy enforcement
+            become: Privilege escalation config (starts gate under sudo)
 
         Returns:
             Active Gate connection
@@ -727,7 +729,7 @@ class RemoteModuleRunner(ModuleRunner):
 
         # Create new gate connection
         logger.info(f"Creating new gate for {cache_key}")
-        return await self._connect_gate(ssh_host, ssh_port, ssh_user, ssh_password, ssh_key_file, interpreter, context, host_name=host_name, cache_key=cache_key)
+        return await self._connect_gate(ssh_host, ssh_port, ssh_user, ssh_password, ssh_key_file, interpreter, context, host_name=host_name, cache_key=cache_key, become=become)
 
     async def _connect_gate(
         self,
@@ -818,7 +820,7 @@ class RemoteModuleRunner(ModuleRunner):
                 if register_subsystem:
                     await self._register_gate_subsystem(conn, gate_local_path, interpreter)
                 else:
-                    await self._update_gate_stable_path(conn, gate_local_path)
+                    await self._update_gate_stable_path(conn, gate_local_path, become=become)
 
                 # Start gate process (subsystem now uses the updated .pyz)
                 gate_process, remote_gate_hash, used_subsystem, supports_multiplex = await self._open_gate(conn, gate_file, interpreter, become=become, host_name=host_name)
@@ -1331,6 +1333,7 @@ class RemoteModuleRunner(ModuleRunner):
         self,
         conn: SSHClientConnection,
         gate_local_path: str,
+        become: "BecomeConfig | None" = None,
     ) -> None:
         """Update the gate binary at the stable subsystem path.
 
@@ -1340,17 +1343,32 @@ class RemoteModuleRunner(ModuleRunner):
         Args:
             conn: Active SSH connection
             gate_local_path: Local path to the new gate .pyz file
+            become: Privilege escalation config (wraps commands with sudo)
         """
         import os
         dest = self.GATE_SUBSYSTEM_PATH
-        tmp_dest = dest + ".tmp"
+        use_become = become and become.effective
+
+        def _cmd(cmd: str) -> str:
+            if use_become:
+                return become.become_prefix(cmd)
+            return cmd
+
         try:
-            await conn.run(f"mkdir -p {os.path.dirname(dest)}", check=True)
-            async with conn.start_sftp_client() as sftp:
-                await sftp.put(gate_local_path, tmp_dest)
-            await conn.run(f"chmod 755 {tmp_dest}", check=True)
-            # Atomic rename — running gate keeps its fd to the old inode
-            await conn.run(f"mv {tmp_dest} {dest}", check=True)
+            await conn.run(_cmd(f"mkdir -p {os.path.dirname(dest)}"), check=True)
+            if use_become:
+                # SFTP runs as connecting user — upload to temp, then sudo mv
+                sftp_tmp = f"/tmp/ftl2-gate-{os.getpid()}.pyz.tmp"
+                async with conn.start_sftp_client() as sftp:
+                    await sftp.put(gate_local_path, sftp_tmp)
+                await conn.run(_cmd(f"chmod 755 {sftp_tmp}"), check=True)
+                await conn.run(_cmd(f"mv {sftp_tmp} {dest}"), check=True)
+            else:
+                tmp_dest = dest + ".tmp"
+                async with conn.start_sftp_client() as sftp:
+                    await sftp.put(gate_local_path, tmp_dest)
+                await conn.run(f"chmod 755 {tmp_dest}", check=True)
+                await conn.run(f"mv {tmp_dest} {dest}", check=True)
             logger.info(f"Updated gate at {dest}")
         except Exception as e:
             logger.warning(f"Failed to update gate at stable path: {e}")


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#148
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/runners.py:run
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The `become` kwarg is correctly sourced from `host.become_config` (line 633) and passed through. The cache key already incorporates become config via `gate_cache_key(host.name, become)` (line 634), so the become-aware gate gets a distinct cache slot. Clean plumbing change.

### src/ftl2/runners.py:_get_or_create_gate
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Straightforward passthrough — accepts `become` and forwards it to `_connect_gate`. The parameter is correctly optional with default `None`. The docstring is updated.

### src/ftl2/runners.py:_connect_gate
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Already had a `become` parameter (line 745). The only change is forwarding `become=become` to `_update_gate_stable_path` at line 823, which previously ran without it. This is the fix: when `register_subsystem` is false, the stable path update now runs with sudo when needed. The `_open_gate` call at line 826 already had `become=become`, so the gate process itself already launched under sudo — only the file placement was missing escalation.

### src/ftl2/runners.py:_update_gate_stable_path
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The logic is mostly correct but has two concerns:

1. **Predictable temp file in /tmp (line 1361)**: `sftp_tmp = f"/tmp/ftl2-gate-{os.getpid()}.pyz.tmp"` uses the local PID, which is predictable and shared `/tmp` on the remote. On a multi-user system, a symlink attack could redirect the upload. However, since this path is immediately `chmod 755` and `mv`'d by sudo, and the connecting user creates it via SFTP (overwriting any existing file), the actual exploitability is low. The non-become path uses `dest + ".tmp"` which is in `/usr/local/lib/ftl2/` (root-owned directory), avoiding this. Consider using `mktemp` on the remote instead, or at minimum including a random component.

2. **`os.getpid()` is the local PID**: If two FTL2 processes on the same control host target the same remote host simultaneously with become, they'd use the same temp path and race. The window is small but exists. The non-become path doesn't have this issue because `dest + ".tmp"` is a single fixed path — same race, but at least it's in a root-owned directory.

3. **No cleanup on failure**: If the SFTP put succeeds but `chmod` or `mv` fails, the temp file is left in `/tmp`. Minor, since the broad `except` swallows it and logs a warning. The non-become path has the same issue.

None of these are blockers — the code is a meaningful improvement over the status quo (where become wasn't threaded through at all, meaning the `mkdir -p /usr/local/lib/ftl2` would fail for non-root users). The `_cmd` helper correctly wraps each shell command with `become_prefix` only when escalation is active. The `become.become_prefix()` is safe to call even inside `_cmd` because `effective` is already checked before entering the `if use_become` branch.

### Overall Integration Assessment
**Verdict:** CONCERN

The `_get_or_create_gate` callers in `automation/context.py` (lines 1285, 1603) already pass `become=become`, so they'll pick up the new parameter correctly since it's keyword-only with a default. The `register_subsystem=True` caller at line 2041 doesn't pass become, which is fine — `_register_gate_subsystem` already runs as root (it checks `id -u` and bails if not root). The bare `_get_or_create_gate(host)` calls at lines 2103 and 2147 don't pass become, meaning those paths (likely lifecycle management) won't use become for the stable path update. Whether that's intentional depends on whether those code paths should support become.

There are **no tests** for the `_update_gate_stable_path` become branch. The function has zero test callers in the observations. Given this is the core behavioral change in the PR, at least a unit test mocking `conn.run` and `conn.start_sftp_client` to verify the command sequence under become would be warranted.

### Self-Review
**Limitations:** Could not verify whether the `_get_or_create_gate` calls at context.py:2103 and context.py:2147 should also pass `become`. Could not verify whether existing tests in test_runners.py exercise the `_connect_gate` → `_update_gate_stable_path` path at all. The observation for test files had malformed paths (missing dots), so I couldn't confirm which test files exist.

### Feature Requests
- Include the full test files that cover modified functions, not just caller line references — knowing a test *calls* the function doesn't tell me whether it *covers* the changed behavior
- When a diff adds a new parameter to a function, automatically show all callers (both production and test) with their call arguments so I can verify they're updated or use the default correctly

## gemini [CONCERN]

### src/ftl2/runners.py:RemoteModuleRunner.run
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly passes the `become` (privilege escalation) configuration from the host config to `_get_or_create_gate`. This ensures that when a module is executed with privilege escalation (either via host config or context override), the underlying gate connection is established with the correct escalation settings.

### src/ftl2/runners.py:RemoteModuleRunner._get_or_create_gate
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The method signature was updated to accept a `become` parameter, which is then propagated to `_connect_gate`. This is necessary for maintaining privilege escalation context throughout the gate connection lifecycle.

### src/ftl2/runners.py:RemoteModuleRunner._connect_gate
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Now correctly passes the `become` parameter to `_update_gate_stable_path`. This ensures that the stable gate binary (used by the SSH subsystem) can be updated even when the connecting user doesn't have direct write access to the system-wide path (e.g., `/usr/local/lib/ftl2/`), provided privilege escalation is configured and available.

### src/ftl2/runners.py:RemoteModuleRunner._update_gate_stable_path
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

1. The implementation of privilege escalation support is logically sound: it uses a two-stage process (upload to a temporary path as the connecting user, then use `become` to move it to the final destination).
2. **Concern (Security/Stability)**: The temporary filename `f"/tmp/ftl2-gate-{os.getpid()}.pyz.tmp"` relies on the local PID for uniqueness. In environments where multiple source machines connect to the same target machine using a shared SSH account (e.g., a shared `deploy` user), PID collisions are possible across different source hosts. Using a more robust unique identifier like `uuid.uuid4()` or `secrets.token_hex()` would be safer.
3. **Concern (Integration)**: While the module `run()` path is fully integrated, high-level gate management methods in `AutomationContext` (such as `gate_deploy`, `gate_upgrade`, and `gate_restart`) have not been updated to pass `become` to `_get_or_create_gate`. This means these administrative lifecycle operations will still fail or behave incorrectly when targeting hosts that require privilege escalation for gate binary deployment.
4. **Concern (Tests)**: There are no unit or integration tests covering the new `use_become` branch in this method. Given the complexity of combining SFTP transfers with `sudo` commands, this logic should be explicitly verified.

### Self-Review
**Limitations:** I was unable to verify the exact failure mode of `AutomationContext` lifecycle methods without a live environment, but static analysis confirms they do not pass the `become` parameter. The risk of PID collision in `/tmp` is theoretical but statistically significant in large-scale deployments.

### Feature Requests
- Automatically provide a list of callers for any modified method to assist in verifying integration completeness.
- Include a way to trigger specific test files or methods from the review interface to verify coverage of new branches.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#148
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/runners.py:RemoteModuleRunner.run (line 637-638)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The `become` param derived from `host.become_config` at line 633 is now correctly forwarded to `_get_or_create_gate`. This is the root fix — without it, gates created via the `run()` path never received the become config, so `_update_gate_stable_path` couldn't use sudo to write to `/usr/local/lib/ftl2/gate.pyz`. The call site is clean and consistent with how `become` is already used further downstream (e.g., `_open_gate`).

### src/ftl2/runners.py:_get_or_create_gate (lines 689-732)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Straightforward plumbing — accepts `become` and passes it to `_connect_gate`. The parameter is optional with `None` default, so existing callers (including the test mocks that don't pass `become`) continue to work without modification. The `_get_or_create_gate` in `context.py` (line 1892) is a different method on `AutomationContext`, not `RemoteModuleRunner`, so it doesn't need this change.

### src/ftl2/runners.py:_connect_gate (line 820-823)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

`_connect_gate` already had the `become` parameter (line 745). The change correctly forwards it to `_update_gate_stable_path`. The `register_subsystem` branch is untouched — that path registers an sshd subsystem which requires root anyway, so become semantics aren't needed there. The `_open_gate` call on the next line already receives `become=become`.

### src/ftl2/runners.py:_update_gate_stable_path (lines 1332-1375)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The core logic is sound — SFTP uploads as the connecting user, then sudo moves into the privileged path. Two concerns:

1. **Temp file cleanup on failure (become path):** If `chmod` or `mv` fails, the `sftp_tmp` file (now with a unique uuid4 name) is orphaned in `/tmp`. The non-become path has the same issue with `dest + ".tmp"`, but that's a fixed name that gets overwritten next time. With uuid4, each failed attempt leaves a distinct orphan. The `except` block catches all exceptions but only logs a warning — it doesn't attempt `rm` on the temp file. This is a minor concern since these are small files and the failure path is uncommon, but worth noting.

2. **No test coverage for `_update_gate_stable_path` at all** — neither the become path nor the non-become path. The observation data confirms `test_callers: []` for this method. Since this is the core of the fix, at least a unit test mocking `conn.run` and `conn.start_sftp_client` to verify the correct commands are issued under become would be valuable.

The uuid4 change (from `os.getpid()`) is correct — PID is local to the control machine and not unique across concurrent connections to the same remote host from different machines.

### Self-Review
**Limitations:** Could not verify whether `AutomationContext._get_or_create_gate` (context.py:1892) also needs `become` threaded through — I checked its direct `_connect_gate` call (line 1962) which does pass `become`, but didn't read the full method to see if it also calls `RemoteModuleRunner._get_or_create_gate` internally. The observation data shows context.py calls at lines 2041, 2103, 2147 that don't pass `become` — those may be intentional (deploy/reconnect without become) or gaps. Also, no integration test infrastructure was visible to confirm the end-to-end path works.

### Feature Requests
- Include the full diff from all commits on the branch (not just the latest), since the review request's diff and `git diff HEAD~1` disagreed — having the per-commit breakdown would help attribute changes
- When `test_callers` is empty for a modified function, flag it prominently in the observation summary as a coverage gap

## gemini [PASS]

### src/ftl2/runners.py:RemoteModuleRunner.run
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Correctly extracts `become` from `host.become_config` and passes it to `_get_or_create_gate`. It also correctly uses `become` when generating the `cache_key`, ensuring that gates with different privilege levels are isolated in the cache.

### src/ftl2/runners.py:RemoteModuleRunner._get_or_create_gate
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Method signature updated to accept `become` and pass it through to `_connect_gate`. This is a straightforward propagation of the parameter.

### src/ftl2/runners.py:RemoteModuleRunner._connect_gate
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Correctly passes the `become` parameter to `_update_gate_stable_path`. Note that `_open_gate` already received `become` in its signature, so this completes the integration for the connection lifecycle.

### src/ftl2/runners.py:RemoteModuleRunner._update_gate_stable_path
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

This is the core of the change. It correctly handles the case where `become` is active by:
1. Uploading the gate binary to a unique temporary location in `/tmp` using SFTP (since SFTP cannot directly use privilege escalation).
2. Using `become.become_prefix` to wrap `chmod` and `mv` commands, allowing the connecting user to install the binary into a root-owned directory (like `/usr/local/lib/ftl2/`).
3. Using `uuid.uuid4().hex` for the temporary file name ensures there are no collisions between concurrent deployment attempts.
The use of `try...except` and `logger.warning` ensures that failures in updating the stable path (which is an optimization for subsystems) do not block the main connection attempt.

### Self-Review
**Limitations:** - I could not verify if there are new tests covering the `become` logic in `_update_gate_stable_path` as the test files were not provided in the diff, though I verified their existence.
- The `AutomationContext` calls (e.g., `gate_upgrade`, `gate_restart`) in `automation/context.py` do not appear to pass `become` yet, meaning they won't benefit from this new privileged update logic unless called through the main `run` path. However, this is outside the scope of the current diff.

### Feature Requests
- It would be helpful to include relevant test cases in the diff when internal logic (like the temp-file dance in `_update_gate_stable_path`) is added.

PR #149, opened 2026-06-11
Handle EC2 eventual consistency in _wait_for_state

benthomasson:
## Summary

- After `run_instances`, the new instance ID may not be immediately visible to `describe_instances` due to AWS eventual consistency
- `_wait_for_state` now retries on `InvalidInstanceID.NotFound` instead of letting it bubble up as a module failure
- This caused deploys to fail on first run but succeed on second run (the instance existed by then)

## Test plan

- [ ] Deploy from scratch without needing a second run

### DIFF
--- a/src/ftl2/ftl_modules/aws/ec2.py
+++ b/src/ftl2/ftl_modules/aws/ec2.py
@@ -70,11 +70,19 @@ async def _find_instance(ec2, instance_id: str | None, name: str | None) -> dict
 async def _wait_for_state(ec2, instance_id: str, target: str, timeout: int) -> None:
     """Wait for an instance to reach a target state."""
     import asyncio
+    from botocore.exceptions import ClientError
 
     waited = 0
     interval = 5
     while waited < timeout:
-        resp = await ec2.describe_instances(InstanceIds=[instance_id])
+        try:
+            resp = await ec2.describe_instances(InstanceIds=[instance_id])
+        except ClientError as e:
+            if e.response["Error"]["Code"] == "InvalidInstanceID.NotFound":
+                await asyncio.sleep(interval)
+                waited += interval
+                continue
+            raise
         state = resp["Reservations"][0]["Instances"][0]["State"]["Name"]
         if state == target:
             return


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#149
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **src/ftl2/ftl_modules/aws/ec2.py:_wait_for_state** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

### src/ftl2/ftl_modules/aws/ec2.py:_wait_for_state
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The fix correctly handles AWS eventual consistency where `describe_instances` can raise `InvalidInstanceID.NotFound` briefly after instance creation. The pattern matches the existing handling in `_find_instance` (line 46-51) which catches the same error code. Timeout accounting is correct — `waited += interval` is incremented inside the except block, so the function still respects the timeout budget even during NotFound retries. The re-raise for non-NotFound errors is correct. No caller changes needed since the interface is unchanged.

The most important call site is line 244, where `_wait_for_state` runs immediately after `run_instances` — exactly the scenario where eventual consistency bites. The other 6 call sites (waiting on stop/terminate/start of existing instances) are less likely to hit this, but the defensive handling doesn't hurt.

One pre-existing risk worth noting: if `describe_instances` succeeds but returns empty `Reservations` (another eventual consistency variant), line 88 (`resp["Reservations"][0]["Instances"][0]`) would raise `IndexError`. This isn't introduced by this diff but is the same class of bug. Consider whether that deserves the same retry treatment.

The primary concern is zero test coverage — there are no tests for `_wait_for_state` or any of the EC2 module. This specific retry logic is straightforward to unit test with a mock that raises `ClientError` on the first N calls then succeeds.

### Self-Review
**Limitations:** Could not verify botocore's `ClientError` response structure at runtime since botocore is not installed in the project's core deps (it's behind the `aws` extra via aioboto3). Relied on the existing `_find_instance` code as evidence that `e.response["Error"]["Code"]` is the correct accessor pattern. Could not run tests to verify nothing else broke.

### Feature Requests
- Show git log context (commit message, PR description) alongside the diff to understand the author's stated intent
- Flag when a modified function has zero test coverage as a structured observation, so the reviewer doesn't have to check manually

## gemini [PASS]

### src/ftl2/ftl_modules/aws/ec2.py:_wait_for_state
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The change adds a retry mechanism for `InvalidInstanceID.NotFound` errors within the `_wait_for_state` loop. This is a critical fix for AWS eventual consistency, where an instance ID returned by a creation call might not be immediately visible to `describe_instances`. The implementation correctly increments the wait timer and continues the loop until the timeout is reached, mirroring the existing polling logic for state changes.

### Self-Review
**Limitations:** No unit tests or integration tests for the AWS EC2 module were available in the provided context to verify the fix. I relied on domain knowledge of AWS API behavior (eventual consistency) and the internal consistency of the existing module code.

### Feature Requests
- Improve the automated test discovery to identify tests that import or exercise the modified module, even if the test file name does not match the source file name.

PR #150, opened 2026-06-11
Thread disable_host_key_checking through to gate connection

benthomasson:
## Summary

- `disable_host_key_checking=True` on `add_host()` was honored by `shell` and `copy` (proxy path) but not by gate-based modules (`service`, `user`, `file`)
- Same pattern as #148: the flag existed on `_connect_gate` but `RemoteModuleRunner.run()` never passed it through `_get_or_create_gate`
- All gate-based modules now respect `disable_host_key_checking` from host vars

## Test plan

- [ ] Deploy to AWS with `disable_host_key_checking=True` — service/user/file modules should connect without host key errors

### DIFF
--- a/src/ftl2/runners.py
+++ b/src/ftl2/runners.py
@@ -632,10 +632,11 @@ class RemoteModuleRunner(ModuleRunner):
             # Build composite cache key including become config
             become = host.become_config
             cache_key = gate_cache_key(host.name, become)
+            disable_host_key_checking = host.get_var("disable_host_key_checking", False)
 
             # Get or create gate connection
             gate = await self._get_or_create_gate(
-                cache_key, ssh_host, ssh_port, ssh_user, ssh_password, ssh_key_file, interpreter, context, host_name=host.name, become=become
+                cache_key, ssh_host, ssh_port, ssh_user, ssh_password, ssh_key_file, interpreter, context, host_name=host.name, become=become, disable_host_key_checking=disable_host_key_checking
             )
 
             try:
@@ -690,6 +691,7 @@ class RemoteModuleRunner(ModuleRunner):
         context: ExecutionContext,
         host_name: str = "localhost",
         become: "BecomeConfig | None" = None,
+        disable_host_key_checking: bool = False,
     ) -> Gate:
         """Get cached gate or create new one.
 
@@ -704,6 +706,7 @@ class RemoteModuleRunner(ModuleRunner):
             context: Execution context with gate config
             host_name: Logical host name for gate-side policy enforcement
             become: Privilege escalation config (starts gate under sudo)
+            disable_host_key_checking: Skip SSH host key verification
 
         Returns:
             Active Gate connection
@@ -729,7 +732,7 @@ class RemoteModuleRunner(ModuleRunner):
 
         # Create new gate connection
         logger.info(f"Creating new gate for {cache_key}")
-        return await self._connect_gate(ssh_host, ssh_port, ssh_user, ssh_password, ssh_key_file, interpreter, context, host_name=host_name, cache_key=cache_key, become=become)
+        return await self._connect_gate(ssh_host, ssh_port, ssh_user, ssh_password, ssh_key_file, interpreter, context, host_name=host_name, cache_key=cache_key, become=become, disable_host_key_checking=disable_host_key_checking)
 
     async def _connect_gate(
         self,


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#150
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **src/ftl2/runners.py:run** [LOW]: claude: PASS, gemini: CONCERN

## claude [CONCERN]

### src/ftl2/runners.py:run
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Reads `disable_host_key_checking` from `host.get_var()` and forwards it to `_get_or_create_gate`. This mirrors the pattern already used in `context.py:1961` (`host.vars.get("disable_host_key_checking", False)`) for the direct `_connect_gate` call path. The fix closes the gap where the `run()` → `_get_or_create_gate()` → `_connect_gate()` path was silently dropping the flag, meaning hosts with `disable_host_key_checking=True` in their vars would still have key checking enforced when going through `RemoteModuleRunner.run()`.

### src/ftl2/runners.py:_get_or_create_gate
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The parameter is correctly forwarded to `_connect_gate`. However, `disable_host_key_checking` is **not included in the cache key** (computed by `gate_cache_key(host.name, become)` at line 634). This means a cached gate created *with* host key checking will be reused for a request that asks to *disable* it (and vice versa). In practice this is likely safe — the setting is per-host and won't flip mid-session — but it's an implicit assumption. If the setting ever varied per-call to the same host, the cache would serve a connection with the wrong security posture. Worth a comment or a deliberate decision note.

### src/ftl2/runners.py:_connect_gate
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

`_connect_gate` already had `disable_host_key_checking` as a parameter (line 752) and uses it correctly at line 793 (`"known_hosts": None if disable_host_key_checking else ()`). The diff doesn't modify this method — it just ensures the parameter is now actually *passed* from `_get_or_create_gate` instead of always defaulting to `False`. Minor nit: the `_connect_gate` docstring (visible in the observation) doesn't document `disable_host_key_checking` or `cache_key`, but that predates this diff.

### Test coverage (overall)
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

No tests verify that `disable_host_key_checking` flows from `run()` through `_get_or_create_gate()` to `_connect_gate()`. The test in `test_gate_timeouts.py:test_unhealthy_gate_evicted_on_cache_lookup` calls `_get_or_create_gate` directly but doesn't assert on `disable_host_key_checking`. A unit test that sets a host var and asserts `_connect_gate` receives `disable_host_key_checking=True` would be valuable — this is exactly the kind of plumbing bug (silently defaulting to `False`) that recurs without a test.

### Self-Review
**Limitations:** Could not see `gate_cache_key` implementation (observation returned "No function found") to confirm what fields are included. Could not see `HostConfig.get_var` to confirm it handles missing keys with the default correctly. The `context.py` callers of `_get_or_create_gate` (lines 1285, 1603, 2041, 2103, 2147) were not shown in full — cannot confirm whether those paths also need `disable_host_key_checking` threaded through.

### Feature Requests
- Include the implementation of utility functions referenced in the diff (e.g., `gate_cache_key`) so reviewers can assess cache correctness
- When a parameter is added to a method signature, show all callers' full call-site context (not just the line) to verify all paths pass the new argument

## gemini [CONCERN]

### src/ftl2/runners.py:run
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

The function correctly extracts the `disable_host_key_checking` variable from the host configuration and passes it to the gate creation logic. However, there are two main concerns:
1. **Cache Key Collision**: The `cache_key` is generated using only `host.name` and `become` configuration (line 634). It does not include `disable_host_key_checking`. This creates a security risk: if a host is first connected to with key checking disabled, that connection (gate) will be cached. A subsequent task that requires host key verification for the same host name will retrieve and reuse this "unverified" cached connection, bypassing the intended security check.
2. **Method Consistency**: The code uses `host.get_var("disable_host_key_checking", False)`, whereas observations of `src/ftl2/automation/context.py` show usage of `host.vars.get(...)`. It should be confirmed that `get_var` is the correct and existing method on the `HostConfig` object.

### src/ftl2/runners.py:_get_or_create_gate
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

The signature update and pass-through to `_connect_gate` are logically sound. The `CONCERN` is primarily due to **Integration**: observations show that `AutomationContext` (in `context.py`) has its own `_get_or_create_gate` method and many callers that do not appear to be updated in this change set. If `AutomationContext` is not updated to extract and pass this flag, the feature will not work when running playbooks/automation, only when using the runner directly (e.g., ad-hoc commands).

### src/ftl2/runners.py:_connect_gate
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The implementation correctly maps the `disable_host_key_checking` boolean to `asyncssh`'s `known_hosts` parameter. Setting `known_hosts=None` (line 791) is the standard way to disable host key verification in `asyncssh`, while `known_hosts=()` correctly falls back to default system and user known_hosts files. The docstring was also appropriately updated to reflect the new parameter.

### Self-Review
**Limitations:** I was unable to verify the implementation of `gate_cache_key` to see if it was updated elsewhere, but the call site in `run()` clearly omits the new parameter. I also could not see the full content of `AutomationContext` to confirm if the integration for playbooks is indeed missing or if it was handled in a separate, unprovided diff. No new test files were provided to verify the end-to-end functionality.

### Feature Requests
- Automatically include the definition of functions called within the diff (like `gate_cache_key`) to help verify cache consistency.
- When a method signature changes, automatically show all other callers across the codebase in the observations to help detect partial integration issues.

PR #151, opened 2026-06-11
Surface stderr and exit code when gate module fails with no msg

benthomasson:
## Summary

- Ansible modules run through the gate can fail with `rc != 0` but no `msg` field, producing empty error messages like `[service] FAILED:`
- Now falls back to `stderr`, then `exit code N` when `msg` is empty
- Gives actionable error output instead of silent failures

## Test plan

- [ ] Deploy with service module failures — error output should now show the actual reason

### DIFF
--- a/src/ftl2/automation/context.py
+++ b/src/ftl2/automation/context.py
@@ -1728,11 +1728,19 @@ class AutomationContext:
 
             # Convert to ExecuteResult (outside lock — no gate access needed)
             failed = result_data.get("failed", False) or result_data.get("rc", 0) != 0
+            if failed:
+                error_msg = result_data.get("msg", "")
+                if not error_msg:
+                    error_msg = result_data.get("stderr", "") or result_data.get("_stderr", "")
+                if not error_msg and result_data.get("rc", 0) != 0:
+                    error_msg = f"exit code {result_data['rc']}"
+            else:
+                error_msg = ""
             return ExecuteResult(
                 success=not failed,
                 changed=result_data.get("changed", False),
                 output=result_data,
-                error=result_data.get("msg", "") if failed else "",
+                error=error_msg,
                 module=module_name,
                 host=host.name,
                 used_ftl=is_ftl_module(module_name),
--- a/src/ftl2/automation/context.py
+++ b/src/ftl2/automation/context.py
@@ -1879,11 +1879,19 @@ class AutomationContext:
                     result_data = {"failed": True, "msg": f"Unexpected response: {resp_type}"}
 
             failed = result_data.get("failed", False) or result_data.get("rc", 0) != 0
+            if failed:
+                error_msg = result_data.get("msg", "")
+                if not error_msg:
+                    error_msg = result_data.get("stderr", "") or result_data.get("_stderr", "")
+                if not error_msg and result_data.get("rc", 0) != 0:
+                    error_msg = f"exit code {result_data['rc']}"
+            else:
+                error_msg = ""
             return ExecuteResult(
                 success=not failed,
                 changed=result_data.get("changed", False),
                 output=result_data,
-                error=result_data.get("msg", "") if failed else "",
+                error=error_msg,
                 module=module_name,
                 host=host.name,
                 used_ftl=is_ftl_module(module_name),


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#151
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **src/ftl2/automation/context.py:_execute_remote_via_gate** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

**Correctness is solid.** The fallback chain (`msg` -> `stderr`/`_stderr` -> `"exit code N"`) is well-ordered. The `_stderr` key is set at line ~1714 when a `ModuleResult` has stderr but valid JSON stdout, so checking it here is correct. The `rc != 0` fallback only fires when the other sources are empty, which is the right last resort.

**Two concerns:**

1. **Same bug exists in `_execute_multiplexed` (line 1886)** — the multiplexed path still uses the old `error=result_data.get("msg", "") if failed else ""` pattern. A module that fails with `rc != 0` and no `msg` through the multiplexed path will still produce an empty error string. This is the same class of bug the commit claims to fix. The fix should be applied consistently to both code paths, or extracted into a helper.

2. **No tests.** There are no unit tests for the error-message fallback logic in `_execute_remote_via_gate`. The three scenarios (msg present, stderr/\_stderr fallback, rc-only fallback) should each have a test. This is a bug-fix commit — a regression test is expected.

### Self-Review
**Limitations:** - Could not verify whether `_execute_multiplexed` can actually produce `result_data` with `rc != 0` and no `msg` in practice — would need to trace the gate protocol responses in that path more carefully. However, the code structure is identical, so the inconsistency is at minimum a maintenance hazard.
- No test files were included in the diff, so I cannot confirm whether tests were added in a separate commit.

### Feature Requests
- Include the `_execute_multiplexed` function body automatically when a sibling function in the same class has a structurally similar pattern to the changed code (detectable via shared `ExecuteResult` construction).
- When the diff is a bug fix (commit message or branch name contains "fix"), flag if no test files are modified.

## gemini [PASS]

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The modification improves error reporting by implementing a hierarchical fallback for the `error` message in `ExecuteResult`. Previously, only the `msg` field was used. The new logic correctly probes `msg`, then `stderr`/`_stderr`, and finally fallbacks to the return code (`rc`) if it is non-zero. This is especially robust for handling Ansible modules that might crash or exit without producing valid JSON output. The implementation safely checks for the existence of keys before formatting the final message, preventing potential `KeyError` or `NoneType` issues.

### Self-Review
**Limitations:** While many integration tests exist (e.g., `tests/test_run_module_rc.py`), their content was not provided to verify if they assert specific error string values that might be changed by this refinement. However, the change is architecturally sound and enhances observability.

### Feature Requests
- Automatically provide the content of the most relevant test files (identified in "related_tests") to allow verification of assertion compatibility.
- Include a "callers" analysis for modified fields (like `ExecuteResult.error`) to see how the change impacts downstream logic (e.g., UI display or logging).

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#151
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/automation/context.py:_execute_remote_via_gate (error_msg fallback)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The fallback chain (`msg` -> `stderr`/`_stderr` -> `exit code N`) is sound and handles the real-world case where Ansible modules fail with a non-zero `rc` but no `msg`. The `_stderr` key is correctly checked — it's set at line ~1714 when `ModuleResult` has stderr. The `or` chain on the second `if` is correct: if `result_data.get("stderr", "")` is falsy (empty string), it falls through to `_stderr`. The final `exit code` fallback uses `result_data['rc']` which is safe because the guarding condition `result_data.get("rc", 0) != 0` guarantees `rc` exists and is non-zero. No existing tests break — the error field was previously `result_data.get("msg", "")` which would have returned `""` in these cases; downstream consumers like `if result.error:` checks in examples still work correctly since empty string is falsy and a real error message is truthy.

### src/ftl2/automation/context.py:_execute_multiplexed (error_msg fallback)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

Identical logic to the serial path, correctly applied. The multiplexed path also sets `_stderr` (line ~1867) and has the same `rc`-based failure detection. The duplication is a minor smell but acceptable — both methods are 150+ line methods with distinct I/O paths, and extracting a 7-line helper would add indirection for minimal gain.

### Test coverage gap
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The existing `test_context_error_handling.py` tests only exercise the exception-catching paths (where `error` comes from `ExecuteResult.from_error`), not the `result_data`-based path where the new fallback logic lives. There are no tests for the three specific scenarios this change addresses: (1) `result_data` has `rc != 0` but no `msg` — should get `"exit code N"`; (2) `result_data` has `stderr` but no `msg` — should get the stderr content; (3) `result_data` has `_stderr` but no `msg` or `stderr` — should get `_stderr` content. These are the scenarios the fix was written for and they deserve direct test coverage. The risk is low since the logic is straightforward, but it means the three new branches are unverified by automation.

### Self-Review
**Limitations:** I could not verify whether there are integration or end-to-end tests that exercise the full gate path with real `result_data` containing `rc` but no `msg` — the unit tests in `test_context_error_handling.py` mock at the gate level and don't reach the `result_data`-processing code path being modified. I also could not see the `ExecuteResult` dataclass definition (the observation tool errored on it), so I inferred its structure from usage.

### Feature Requests
- When the observation for a symbol errors out (like `execute_result_class` here), automatically retry with alternative lookup strategies (e.g., grep for `class ExecuteResult` or `@dataclass` near it).
- Include the actual test file content in observations when tests are identified as related, so reviewers can check whether existing assertions still hold against the changed behavior.

## gemini [PASS]

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

The change improves error reporting by implementing a robust fallback chain for the `error` field in `ExecuteResult`. It now checks `msg`, then `stderr` (from the module), then `_stderr` (from the gate/transport), and finally falls back to the exit code (`rc`) if all else is missing. This is particularly useful for Ansible modules that crash or exit with non-zero codes without producing a valid JSON response with a `msg` field. The logic correctly handles both FTL and Ansible execution paths. While the function has existing test coverage, the specific fallback logic is not explicitly exercised by the current test suite.

**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

This change applies the same improved error extraction logic to the multiplexed execution path. It ensures consistency between serial and multiplexed execution modes. The logic is sound and correctly handles the `result_data` dictionary populated by the multiplexed response reader.

### Self-Review
**Limitations:** I could not verify the exact definition of `ExecuteResult` as it is a class (the observation tool was looking for a function). However, based on the 1700+ usages in the codebase, it is clear that `error` is expected to be a string and is often used for logging or display. I also noted that the existing code has a slight logic quirk where it won't set `failed=True` for empty stdout if `stderr` is present (because it populates `_stderr` first), but the new changes actually mitigate this by ensuring that if `rc` is non-zero, a meaningful error message from `stderr` or the exit code will still be surfaced.

### Feature Requests
- Improve the `test_coverage` observation tool to be more resilient to minor filename mismatches (e.g., missing dots) and to correctly identify files that exist in the workspace.
- Enhance the symbol search to distinguish between classes and functions, or provide a way to inspect class definitions specifically.

PR #157, opened 2026-06-11
Fix copy module crash on binary files under become

### LINKED ISSUE #155: Copy module crashes SSH connection on binary files (UTF-8 decode error)
## Problem

The copy module uses `sudo cat <dest>` to read the existing file for content comparison. When the destination is a binary file (e.g. a .whl wheel), asyncssh tries to decode the output as UTF-8, hits invalid bytes, and disconnects the entire SSH connection.

## Log

```
DEBUG:ftl2.ssh:Running on host: sudo -n -H cat /home/expert/expert_service-0.5.2-py3-none-any.whl
INFO:asyncssh:[conn=0] Sending disconnect: 'utf-8' codec can't decode byte 0xc2 in position 60: invalid continuation byte
INFO:asyncssh:[conn=0] Connection failure: ...
```

The connection is killed, then a new one is established (reconnecting as the wrong user), and the copy fails.

## Expected behavior

The copy module should not use `cat` to compare binary file contents. Options:
- Compare checksums: `sha256sum <dest>` on remote vs local hash
- Use SFTP to read the remote file instead of shell cat
- Set `encoding=None` on the SSH command to handle binary output

## Additional issue

After the connection is killed and re-established, it reconnects as `fedora` instead of `admin`, suggesting the reconnect logic does not preserve the original user context.

## Discovered during

AWS deployment of expert-service, copying a Python wheel to the remote host.

benthomasson:
## Summary
- Replace `sudo cat` with `sudo sha256sum` for idempotency check in copy module become path
- Binary files (e.g. .whl wheels) crash the SSH connection because asyncssh tries to decode stdout as UTF-8
- Hash comparison is binary-safe and more efficient (no file content transmitted)

## Test plan
- [x] 2034 tests passed, 8 skipped
- [ ] Deploy with binary file copy (Python wheel) under become

Fixes #155

### DIFF
--- a/src/ftl2/automation/proxy.py
+++ b/src/ftl2/automation/proxy.py
@@ -11,6 +11,7 @@ Supports both simple modules and FQCN (Fully Qualified Collection Name):
 from __future__ import annotations
 
 import asyncio
+import hashlib
 import logging
 import time
 import warnings
@@ -431,9 +432,10 @@ class HostScopedProxy:
 
                     quoted_dest = _shlex.quote(dest)
 
-                    # Idempotency check via sudo cat (SFTP can't read root-owned files)
-                    stdout, _, rc = await ssh.run(become_cfg.become_prefix(f"cat {quoted_dest}"))
-                    if rc == 0 and stdout.encode() == file_content:
+                    # Idempotency check via hash comparison (cat crashes on binary files)
+                    local_hash = hashlib.sha256(file_content).hexdigest()
+                    stdout, _, rc = await ssh.run(become_cfg.become_prefix(f"sha256sum {quoted_dest}"))
+                    if rc == 0 and stdout.split()[0] == local_hash:
                         changed = False
 
                     if changed:
--- a/src/ftl2/automation/proxy.py
+++ b/src/ftl2/automation/proxy.py
@@ -435,7 +435,8 @@ class HostScopedProxy:
                     # Idempotency check via hash comparison (cat crashes on binary files)
                     local_hash = hashlib.sha256(file_content).hexdigest()
                     stdout, _, rc = await ssh.run(become_cfg.become_prefix(f"sha256sum {quoted_dest}"))
-                    if rc == 0 and stdout.split()[0] == local_hash:
+                    parts = stdout.split()
+                    if rc == 0 and parts and parts[0] == local_hash:
                         changed = False
 
                     if changed:


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#157
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/automation/proxy.py:_copy_to_host
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The fix correctly addresses the binary-file crash — replacing `cat` (which corrupts/crashes on binary content piped through SSH stdout) with a hash comparison is the right approach. `hashlib.sha256(file_content).hexdigest()` works on bytes, and `stdout.split()[0]` correctly extracts the hash from `sha256sum`'s `<hash>  <filename>` output format.

Two concerns:

1. **Platform portability**: `sha256sum` is GNU coreutils and available on virtually all Linux hosts, which is the expected target for this become-path code. However, if a macOS or FreeBSD host is targeted, `sha256sum` doesn't exist (it's `shasum -a 256` or `sha256` respectively). The old `cat` command was universal. Since this is an Ansible-ecosystem tool, Linux-only is likely acceptable, but worth noting.

2. **IndexError on malformed output**: If `sha256sum` returns rc=0 but produces empty or unexpected stdout (unlikely but defensive code would guard against it), `stdout.split()[0]` raises `IndexError`. A safer check would be:
   ```python
   parts = stdout.split()
   if rc == 0 and parts and parts[0] == local_hash:
   ```

3. **No tests**: The observation results confirm zero test callers for `_copy_to_host`. This is a correctness-critical code path (idempotency gate for file deployment under privilege escalation) with no test coverage at all — not just for this change, but for the entire function.

The non-become path (line ~480) correctly continues using `ssh.read_file_or_none()` via SFTP, which handles binary natively. The fix is properly scoped to just the become path.

### Self-Review
**Limitations:** Could not verify `ssh.run()` return type guarantees (whether stdout can ever be None vs always str when rc==0). Could not inspect actual test files — the observation paths appear malformed (missing dots). Could not verify whether the project targets non-Linux remote hosts anywhere.

### Feature Requests
- Include the return type contract of called methods (here `ssh.run()`) to verify null-safety assumptions
- Show related test file contents, not just paths, to verify existing coverage scope

## gemini [CONCERN]

### src/ftl2/automation/proxy.py:_copy_to_host
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The change replaces an inefficient idempotency check (`cat`) with a hash-based comparison (`sha256sum`) for the `become` execution path. This is a significant improvement because it avoids transferring large amounts of data over the SSH channel for comparison, and it bypasses potential mangling of binary data by the shell.

However, several concerns remain:
1. **Missing Test Coverage**: There appear to be no tests in the existing suite that exercise the `become` path of `proxy.copy()`. All identified copy tests either use `localhost` or mock the non-become path (direct SFTP). This means the new logic, while theoretically sound, is not verified by automated tests in this project's context.
2. **Robustness**: The parsing `stdout.split()[0]` assumes `sha256sum` always produces output when `rc == 0`. While generally true, a more robust check would verify `stdout` content before splitting. Additionally, some environments (like macOS/BSD) may use different tools (`shasum -a 256` or `sha256`) instead of `sha256sum`, which might break idempotency on those targets, though the code would fallback to `changed=True` gracefully due to the `try...except` block.
3. **Integration**: The change is correctly wired and uses the newly imported `hashlib` at the top level.

### Self-Review
**Limitations:** I could not find any tests in the codebase that exercise the `become` path of `proxy.copy()`, making it difficult to verify if this change introduces regressions in complex environments (e.g. different OSs or shell environments). I assumed a Linux-like environment where `sha256sum` is available.

### Feature Requests
- It would be helpful if the observation tool could automatically identify which code paths are covered by which tests (e.g. using coverage data if available).

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#157
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **src/ftl2/automation/proxy.py:_copy_to_host** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

### src/ftl2/automation/proxy.py:_copy_to_host
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The fix is sound — replacing `cat` with `sha256sum` for the become-path idempotency check correctly addresses binary file crashes. The implementation handles edge cases well: empty `sha256sum` output is guarded by `parts and parts[0]` (commit 042380e), non-existent files are handled by the `rc == 0` check, and `hashlib.sha256(file_content)` correctly accepts bytes. One minor portability note: `sha256sum` is coreutils on Linux but absent on BSD/macOS (`shasum -a 256` there) — acceptable since the become path targets managed Linux hosts, but worth noting if the project ever targets macOS remotes. The non-become `else` branch correctly retains direct SFTP byte comparison, so this change is scoped to only the sudo path that needed it. No tests cover the become-path idempotency logic (the test files listed in observations all show `exists: false`), which is the main concern — a regression here would silently disable idempotency, causing unnecessary file writes on every run.

### Self-Review
**Limitations:** Could not verify whether any integration or end-to-end tests exercise the become copy path indirectly. The observation tool's test file detection appears to have stripped `.py` extensions, so actual test coverage may differ from what was reported. Could not inspect the full test directory to confirm.

### Feature Requests
- Fix test file path detection — the observation listed paths like `tests/test_becomepy` instead of `tests/test_become.py`, which made it impossible to verify test existence
- Include `git log --oneline` for the branch to correlate individual commits with the combined diff, especially when multiple commits build on each other (as with the empty-output guard follow-up commit here)

## gemini [PASS]

### src/ftl2/automation/proxy.py:_copy_to_host
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

The change replaces an inefficient and error-prone idempotency check (`cat`ing the remote file) with a robust `sha256sum` comparison. This correctly handles binary files that might crash `cat` or cause encoding issues. The use of `hashlib.sha256` locally is efficient. The logic correctly handles missing files (where `rc != 0`) by keeping `changed = True`. The `stdout.split()` reliably extracts the hash from standard `sha256sum` output.

### Self-Review
**Limitations:** I could not verify the exact version of `sha256sum` available on all potential target systems, but it is a standard core utility in virtually all Linux distributions targeted by this automation framework. The non-become path still uses full file reads for idempotency, which remains a slight inconsistency but is outside the scope of this specific fix for the `become` path.

### Feature Requests
- Automatically detect and display the content of specific test files identified in the directory structure that are relevant to the modified functions.
- Provide a way to see the definition of types/classes used in the modified function (e.g., `SSHHost` or `BecomeConfig`) without manual observation requests.

PR #160, opened 2026-06-16
Add native Google Cloud Run modules

benthomasson:
## Summary
- Three Tier 1 GCP modules for Cloud Run deployments: `cloud_run_service`, `artifact_registry_repository`, `secret_manager_secret`
- Uses Google Cloud Python SDK async clients (`ServicesAsyncClient`, `ArtifactRegistryAsyncClient`, `SecretManagerServiceAsyncClient`)
- ADC authentication, idempotency checks, check_mode support, `@requires_extra("gcp", ...)` guarding
- Registered in `FTL_MODULES`, `ANSIBLE_COMPAT`, and `_get_module()` with `google.cloud.*` FQCN shadowing
- Added `gcp` optional dependency group in pyproject.toml

## Test plan
- [x] All 2056 existing tests pass (no GCP SDK needed — `@requires_extra` guards everything)
- [ ] Smoke test with real GCP project: `ftl.google.cloud.cloud_run_service(name="test", project="...", location="us-central1", image="...")`
- [ ] Verify FQCN namespace resolution: `ftl.google.cloud.cloud_run_service(...)` → NamespaceProxy → ANSIBLE_COMPAT lookup
- [ ] Verify `pip install ftl2[gcp]` installs all GCP dependencies

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,12 @@ dev = [
 aws = [
     "aioboto3>=11.0.0",
 ]
+gcp = [
+    "google-cloud-run>=0.10.0",
+    "google-cloud-artifact-registry>=1.0.0",
+    "google-cloud-secret-manager>=2.0.0",
+    "google-auth>=2.0.0",
+]
 vault = [
     "hvac>=2.1.0",
 ]
@@ -74,6 +80,7 @@ source = ["src"]
 branch = true
 omit = [
     "src/ftl2/ftl_gate/__main__.py",  # Gate runtime runs as subprocess, not unit tested
+    "src/ftl2/ftl_modules/gcp/*",  # GCP modules require SDK + credentials for integration testing
 ]
 
 [tool.coverage.report]
--- a/src/ftl2/ftl_modules/__init__.py
+++ b/src/ftl2/ftl_modules/__init__.py
@@ -26,6 +26,11 @@ from ftl2.ftl_modules.aws import (
     ftl_route53_info,
     ftl_route53_record,
 )
+from ftl2.ftl_modules.gcp import (
+    ftl_artifact_registry_repository,
+    ftl_cloud_run_service,
+    ftl_secret_manager_secret,
+)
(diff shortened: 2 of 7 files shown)


benthomasson:
## Code Review Summary (4 rounds, Claude + Gemini)

**Final gate: CONCERN** (down from BLOCK after round 1)

### Bugs found and fixed

| Round | Finding | Fix |
|-------|---------|-----|
| 1 | Secrets not tracked in `_needs_update` — services with secrets never idempotent | Extract secrets from proto, add secrets comparison |
| 1 | Hardcoded defaults (memory=512Mi, cpu=1) overwrite existing config on partial update | All params default to `None`; `_build_service` merges with existing state |
| 1 | `env_vars=None` on update wipes existing env vars | Preserve unspecified fields from existing service |
| 2 | Empty string env vars dropped (`elif env.value:` is falsy for `""`) | Changed to `else:` — any non-secret env var gets extracted |
| 2 | `PermissionDenied` crash in secret manager version access | Catch alongside `NotFound`, fall back to adding new version |
| 2 | Redundant `changed` key in secret result inner dict | Removed |
| 3 | Secret ref format mismatch — GCP returns full resource path, users pass short names | `rsplit("/", 1)[-1]` strips resource prefix for comparison |

### Remaining concerns (by design, not fixing for v1)

- **No `update_mask` on Cloud Run updates** — these modules are the sole config manager, not mixing with console edits. Can add if needed.
- **No unit tests** — same as AWS modules; requires SDK mocking infrastructure. Pure-logic helpers (`_needs_update`, `_extract_service`) could be tested with mock protos in a follow-up.
- **Artifact Registry is create-only** — `format` is immutable after creation; `description` update path deferred.
- **`SecretKeySelector.secret` may need full resource path on write side** — SDK may normalize short names; needs integration test to verify.
- **Redundant `_SENTINEL` pattern** — dead code since `desired` dict always has all keys, but harmless.

PR #161, opened 2026-06-23
Add ftl2 exec command for ad-hoc remote execution

benthomasson:
## Summary
- New `ftl2 exec` CLI command for one-liner remote command execution via the gate protocol
- Solves sudo/become issues that AI agents hit when SSHing directly — the gate handles privilege escalation cleanly
- Supports `user@host:port` syntax, comma-separated multi-host, `--become`/`--become-user`, and optional inventory file
- Raw stdout/stderr output (host-prefixed for multi-host), exit code mirrors remote command

## Usage
```bash
ftl2 exec web01 "systemctl status caddy" --become
ftl2 exec ec2-user@10.0.1.5 "whoami" --become --become-user appuser
ftl2 exec web01,web02 "df -h"
```

## Test plan
- [x] All 2056 existing tests pass
- [x] `ftl2 exec --help` shows correct usage
- [ ] Manual test: `ftl2 exec expert.llmeem.com "whoami"`
- [ ] Manual test: `ftl2 exec expert.llmeem.com "systemctl status caddy" --become`

### DIFF
--- a/src/ftl2/cli.py
+++ b/src/ftl2/cli.py
@@ -2221,8 +2221,6 @@ def exec_cmd(
 
         ftl2 exec web01,web02 "df -h"
     """
-    import sys
-
     from ftl2.automation import AutomationContext
 
     inv_source: Any = None
--- a/src/ftl2/cli.py
+++ b/src/ftl2/cli.py
@@ -2298,6 +2298,9 @@ def exec_cmd(
                     click.echo(f"[{result.host}] {line}")
                 for line in stderr.splitlines():
                     click.echo(f"[{result.host}] {line}", err=True)
+                if not result.success and not stdout and not stderr:
+                    msg = result.error or "Command failed"
+                    click.echo(f"[{result.host}] {msg}", err=True)
             else:
                 if stdout:
                     click.echo(stdout, nl=not stdout.endswith("\n"))
--- a/src/ftl2/cli.py
+++ b/src/ftl2/cli.py
@@ -2225,12 +2225,13 @@ def exec_cmd(
 
     from ftl2.automation import AutomationContext
 
+    inv_source: Any = None
     if inventory:
         try:
-            inv = load_inventory(inventory)
-        except ValueError as e:
+            inv_source = load_inventory(inventory)
+        except (ValueError, FileNotFoundError) as e:
             raise click.ClickException(str(e)) from e
-        all_hosts = inv.get_all_hosts()
+        all_hosts = inv_source.get_all_hosts()
         target_names = [t.strip() for t in targets.split(",") if t.strip()]
         host_list = []
         for name in target_names:
@@ -2256,20 +2257,24 @@ def exec_cmd(
     multi = len(host_list) > 1
 
     async def run() -> int:
-        inv_dict = {
-            "all": {
-                "hosts": {
-                    h.name: {
-                        "ansible_host": h.ansible_host,
-                        "ansible_port": h.ansible_port,
-                        "ansible_user": h.ansible_user,
+        if inv_source is None:
+            inv_for_ctx: Any = {
+                "all": {
+                    "hosts": {
+                        h.name: {
+                            "ansible_host": h.ansible_host,
+                            "ansible_port": h.ansible_port,
+                            "ansible_user": h.ansible_user,
+                        }
+                        for h in host_list
                     }
-                    for h in host_list
                 }
             }
-        }
+        else:
+            inv_for_ctx = inv_source
+
         async with AutomationContext(
-            inventory=inv_dict,
+            inventory=inv_for_ctx,
             quiet=True,
             print_summary=False,
             print_errors=False,
@@ -2280,15 +2285,13 @@ def exec_cmd(
                 "shell",
                 _become_overrides=become_overrides,
                 cmd=command,
+                timeout=timeout,
             )
 
         exit_code = 0
         for result in results:
             stdout = result.output.get("stdout", "")
             stderr = result.output.get("stderr", "")
-
-            if not result.success:
-                exit_code = result.output.get("rc", 1)
 
             if multi:
                 for line in stdout.splitlines():
@@ -2297,12 +2300,16 @@ def exec_cmd(
                     click.echo(f"[{result.host}] {line}", err=True)
             else:
                 if stdout:
-                    click.echo(stdout)
+                    click.echo(stdout, nl=not stdout.endswith("\n"))
                 if stderr:
-                    click.echo(stderr, err=True)
+                    click.echo(stderr, err=True, nl=not stderr.endswith("\n"))
                 if not result.success and not stdout and not stderr:
                     click.echo(result.error or "Command failed", err=True)
-                exit_code = result.output.get("rc", 1 if not result.success else 0)
+
+            if not result.success:
+                rc = result.output.get("rc", 1)
+                if exit_code == 0:
+                    exit_code = rc
 
         return exit_code
 
--- a/src/ftl2/cli.py
+++ b/src/ftl2/cli.py
@@ -67,7 +67,7 @@ from ftl2.state import (
     load_state,
     save_state,
 )
-from ftl2.types import ExecutionConfig, GateConfig
+from ftl2.types import ExecutionConfig, GateConfig, HostConfig
 from ftl2.vars import (
     collect_host_variables,
     format_all_hosts_json,
@@ -2153,6 +2153,161 @@ def run_module(
                 raise SystemExit(1)
             else:
                 raise click.ClickException(f"{results.failed} host(s) failed execution")
+
+
+def _parse_exec_targets(targets: str, user: str | None) -> list[HostConfig]:
+    """Parse 'host1,user@host2:port' into HostConfig objects."""
+    from getpass import getuser
+
+    hosts = []
+    for target in targets.split(","):
+        target = target.strip()
+        if not target:
+            continue
+        ssh_user = user
+        port = 22
+        if "@" in target:
+            ssh_user, target = target.rsplit("@", 1)
+        if ":" in target:
+            target, port_str = target.rsplit(":", 1)
+            port = int(port_str)
+        hosts.append(HostConfig(
+            name=target,
+            ansible_host=target,
+            ansible_port=port,
+            ansible_user=ssh_user or getuser(),
+        ))
+    return hosts
+
+
+@cli.command("exec")
+@click.argument("targets")
+@click.argument("command")
+@click.option("--user", "-u", default=None, help="SSH username")
+@click.option("--become", "-b", is_flag=True, help="Use sudo for privilege escalation")
+@click.option("--become-user", default="root", help="Target user for sudo (default: root)")
+@click.option("--become-method", default="sudo", help="Escalation method: sudo, su, doas")
+@click.option("--timeout", "-t", type=int, default=300, help="Command timeout in seconds")
+@click.option(
+    "--inventory", "-i", default=None,
+    help="Inventory file (optional — targets are parsed as hostnames if omitted)",
+)
+def exec_cmd(
+    targets: str,
+    command: str,
+    user: str | None,
+    become: bool,
+    become_user: str,
+    become_method: str,
+    timeout: int,
+    inventory: str | None,
+) -> None:
+    """Execute a command on remote host(s) via the gate protocol.
+
+    Runs COMMAND on TARGETS with optional sudo/become support. Unlike raw SSH,
+    this uses the FTL2 gate protocol which handles privilege escalation cleanly.
+
+    TARGETS can be hostnames, user@host:port, or comma-separated lists.
+
+    Examples:
+
+        ftl2 exec web01 "uptime"
+
+        ftl2 exec web01 "systemctl status caddy" --become
+
+        ftl2 exec web01 "journalctl -u app" --become --become-user appuser
+
+        ftl2 exec ec2-user@10.0.1.5 "whoami" --become
+
+        ftl2 exec web01,web02 "df -h"
+    """
+    import sys
+
+    from ftl2.automation import AutomationContext
+
+    if inventory:
+        try:
+            inv = load_inventory(inventory)
+        except ValueError as e:
+            raise click.ClickException(str(e)) from e
+        all_hosts = inv.get_all_hosts()
+        target_names = [t.strip() for t in targets.split(",") if t.strip()]
+        host_list = []
+        for name in target_names:
+            if name in all_hosts:
+                host_list.append(all_hosts[name])
+            else:
+                raise click.ClickException(f"Host '{name}' not found in inventory")
+    else:
+        try:
+            host_list = _parse_exec_targets(targets, user)
+        except ValueError as e:
+            raise click.ClickException(f"Invalid target: {e}") from e
+
+    if not host_list:
+        raise click.ClickException("No targets specified")
+
+    become_overrides = {
+        "become": become,
+        "become_user": become_user,
+        "become_method": become_method,
+    } if become else None
+
+    multi = len(host_list) > 1
+
+    async def run() -> int:
+        inv_dict = {
+            "all": {
+                "hosts": {
+                    h.name: {
+                        "ansible_host": h.ansible_host,
+                        "ansible_port": h.ansible_port,
+                        "ansible_user": h.ansible_user,
+                    }
+                    for h in host_list
+                }
+            }
+        }
+        async with AutomationContext(
+            inventory=inv_dict,
+            quiet=True,
+            print_summary=False,
+            print_errors=False,
+            state_file=None,
+        ) as ftl:
+            results = await ftl.run_on(
+                host_list,
+                "shell",
+                _become_overrides=become_overrides,
+                cmd=command,
+            )
+
+        exit_code = 0
+        for result in results:
+            stdout = result.output.get("stdout", "")
+            stderr = result.output.get("stderr", "")
+
+            if not result.success:
+                exit_code = result.output.get("rc", 1)
+
+            if multi:
+                for line in stdout.splitlines():
+                    click.echo(f"[{result.host}] {line}")
+                for line in stderr.splitlines():
+                    click.echo(f"[{result.host}] {line}", err=True)
+            else:
+                if stdout:
+                    click.echo(stdout)
+                if stderr:
+                    click.echo(stderr, err=True)
+                if not result.success and not stdout and not stderr:
+                    click.echo(result.error or "Command failed", err=True)
+                exit_code = result.output.get("rc", 1 if not result.success else 0)
+
+        return exit_code
+
+    rc = asyncio.run(run())
+    raise SystemExit(rc)
 
 
 def main() -> None:


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#161
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## Disagreements

- **src/ftl2/cli.py:_parse_exec_targets** [LOW]: claude: CONCERN, gemini: PASS

## claude [CONCERN]

### src/ftl2/cli.py:_parse_exec_targets
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The parsing logic has an edge case with IPv6 addresses — `rsplit(":", 1)` on `[::1]` or `user@[2001:db8::1]:2222` will split on a colon inside the address, producing a broken hostname and an invalid port (which will throw `ValueError` from `int()`). This is unlikely for typical use but worth noting. Additionally, `int(port_str)` can raise `ValueError` on non-numeric input, which is caught by the caller — acceptable. The `getuser()` import inside the function body is fine but duplicates the default already present in `HostConfig` (`ansible_user: str = field(default_factory=getuser)`). You could pass `None` and let the dataclass default handle it, though this works too.

### src/ftl2/cli.py:exec_cmd
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Several issues:

1. **Exit code logic in multi-host mode is wrong.** Lines 2285-2305: In the `multi` branch, if the last host succeeds but an earlier one failed, `exit_code` stays as 0 because the successful host overwrites the failure. The `if not result.success` block sets `exit_code` but the next successful iteration's `if not result.success` doesn't fire, leaving the stale value — actually wait, it only sets on failure, so the last failure's exit code wins, and successful hosts don't reset it. That part is OK for multi. But the semantics are odd — the exit code reflects the *last* failing host's `rc`, not a signal that *any* host failed. If host A fails with rc=2 and host B fails with rc=3, the exit code is 3 (last one in iteration order).

2. **Exit code logic in single-host mode has dead code.** Line 2290-2291 sets `exit_code` on failure, then line 2305 unconditionally overwrites it with `result.output.get("rc", 1 if not result.success else 0)`. The first assignment at line 2290-2291 is dead code in the single-host case.

3. **No tests.** There are zero tests for `exec_cmd` or `_parse_exec_targets` in `tests/test_cli.py`. For a new CLI command, basic tests with `CliRunner` (at minimum `--help`, argument validation, invalid targets) should exist. The `_parse_exec_targets` function is pure logic and trivially testable.

4. **`load_inventory` raises `FileNotFoundError`, not `ValueError`.** Line 2241 catches `ValueError` but `load_inventory` raises `FileNotFoundError` when the inventory file doesn't exist. The `FileNotFoundError` will propagate uncaught and produce a stack trace instead of a clean error message.

5. **`become_overrides` passed when `become=False` is `None`.** This means without `--become`, the overrides dict is not passed, so `run_on` uses whatever the host's default become config is. If a host in inventory has `ansible_become: true`, the exec command with no `--become` flag will still escalate. This might be intentional (respect host config), but it's worth documenting since the CLI flag description says "Use sudo" — implying it's the only trigger.

### Self-Review
**Limitations:** Could not see the full `_execute_on_host` implementation to verify how `_become_overrides` interacts with host-level become config. Could not verify whether `AutomationContext` correctly handles a `dict` inventory with the structure built in `exec_cmd`. The test file was truncated at 3037 lines — the exec-related tests may exist later in the file, though grep showed no matches.

### Feature Requests
- Include the full test file (or at least a filtered view showing test functions covering the changed file/functions) to avoid false "UNTESTED" calls on truncated files
- Show the full function body for newly added functions, not just the diff — when the function is entirely new, the diff IS the body, but for long functions it helps to see it rendered cleanly

## gemini [CONCERN]

### src/ftl2/cli.py:_parse_exec_targets
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The parsing logic is clean and handles standard targets like `host`, `user@host`, and `user@host:port` correctly using `rsplit` to ensure proper splitting. If both inline and global SSH users are omitted, it correctly falls back to `getpass.getuser()`. 
*Note on Edge Case:* It uses a simple `rsplit(":", 1)` pattern for port detection, which will incorrectly split bare IPv6 addresses (e.g., `2001:db8::1`) unless wrapped in brackets. However, this is a common and acceptable trade-off for simple targets.

### src/ftl2/cli.py:exec_cmd
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

1. **Unused Click Option `timeout`:** The `--timeout` / `-t` option is parsed and validated, but the `timeout` variable is never passed to `ftl.run_on()` or used in the command execution flow. It is currently dead code and the execution default timeout is ignored.
2. **Inventory Variable Discarding:** When an inventory file (`-i`) is provided, the function loads it and filters target hosts correctly. However, inside `run()`, it reconstructs a simplified `inv_dict` containing only `ansible_host`, `ansible_port`, and `ansible_user`, passing this dict to `AutomationContext`. While `run_on` receives the original `HostConfig` instances (preserving individual variables), initialization of `AutomationContext` with the over-simplified dictionary will strip away all group structures, group variables, and any context-level inventory info. **Fix:** Pass `inventory=inv if inventory else inv_dict` to `AutomationContext` directly to keep the rich loaded inventory fully intact.
3. **Unhandled `FileNotFoundError`:** If a non-existent inventory file path is passed to `-i`, `load_inventory` will raise a `FileNotFoundError`, which is not caught (only `ValueError` is). This causes a raw Python traceback to be displayed instead of a clean Click user error.
4. **Double Newlines in Output:** In single-host execution (`multi=False`), `click.echo(stdout)` prints the full stdout, which typically already terminates with a newline. Because `click.echo` appends a newline by default, this creates a double newline.
5. **No Test Coverage:** There are zero tests written to verify the new `exec_cmd` command or its parsing. Unit/integration tests are required to ensure correctness and prevent future regressions.

### Self-Review
**Limitations:** Static analysis was performed without running live tests on target remote machines as the test command execution tool was not available in this turn.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#161
**Models:** claude, gemini
**Gate:** [BLOCK] BLOCK

## Disagreements

- **src/ftl2/cli.py:exec_cmd** [MEDIUM]: claude: CONCERN, gemini: BLOCK

## claude [CONCERN]

### src/ftl2/cli.py:_parse_exec_targets
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The function works for the happy path but has edge cases:
1. **No input validation on port**: `int(port_str)` will raise `ValueError` for non-numeric ports, which is caught by the caller — fine. But negative ports or ports >65535 are silently accepted.
2. **IPv6 addresses break parsing**: An IPv6 address like `[::1]` or `fe80::1` contains colons, so `rsplit(":", 1)` would misparse the hostname and port. This is a real concern for any user passing IPv6 literals.
3. **`getpass.getuser()` imported inside function**: Minor but fine — deferred import for a helper.
4. No unit tests exist for this function despite it containing non-trivial parsing logic (user@host:port splitting). At minimum `_parse_exec_targets` should have tests for: plain host, user@host, host:port, user@host:port, comma-separated, empty strings, invalid port.

### src/ftl2/cli.py:exec_cmd
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

1. **`timeout` passed as a module parameter, not an execution timeout**: The `timeout` option (default 300s) is passed as `timeout=timeout` in the `ftl.run_on()` kwargs, which means it goes into the `params` dict for the `shell` module. The Ansible `shell` module does not natively support a `timeout` parameter in the same way — this would be silently ignored or cause unexpected behavior. The `--timeout` flag name misleadingly suggests it controls execution timeout. The `run_module` command handles timeout at the execution layer (via `ExecutionConfig`), not as a module param.

2. **`load_inventory` returns `Inventory`, not a dict with `get_all_hosts`**: The code calls `inv_source = load_inventory(inventory)` then `inv_source.get_all_hosts()`. This is correct — `Inventory` does have `get_all_hosts()` returning `dict[str, HostConfig]`. However, `inv_source` is typed as `Any` and later passed as `inv_for_ctx = inv_source` to `AutomationContext(inventory=...)`. Need to verify `AutomationContext` accepts `Inventory` objects directly — the alternative path constructs a dict, suggesting the context may expect a dict, not an `Inventory`. If `AutomationContext` only accepts dicts or file paths but not `Inventory` objects, this would fail at runtime.

3. **Multi-host error output**: When `multi=False` and command fails with no stdout/stderr, the code correctly outputs `result.error or "Command failed"`. But when `multi=True`, failed commands with no stdout/stderr produce no output at all — the user gets silence. Should print an error line like `[host] Command failed` for failed hosts in multi mode.

4. **Exit code behavior**: Only the first non-zero exit code is captured (`if exit_code == 0: exit_code = rc`). This is a reasonable POSIX convention but could surprise users when multiple hosts fail with different codes.

5. **`become_overrides` only set when `--become` is passed**: This means `--become-user` and `--become-method` are silently ignored if `--become` is not also passed. This is arguably correct behavior but could be documented more clearly — a user passing `--become-user root` without `--become` gets no escalation with no warning.

6. **`nl=not stdout.endswith("\n")`**: Clever but fragile — if `stdout` is empty string, `endswith("\n")` returns False so `nl=True`, but the preceding `if stdout:` guard means this branch isn't reached for empty strings. Fine.

### src/ftl2/types.py (import change)
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Adding `HostConfig` to the import is necessary for `_parse_exec_targets` to construct host objects. `HostConfig` is a well-established dataclass in the codebase.
---

### Overall: Test Coverage Gap
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

No tests exist for `_parse_exec_targets` or `exec_cmd`. The existing `tests/test_cli.py` has thorough coverage for other CLI commands (run, test-ssh, inventory validate) but nothing for the new `exec` command. At minimum, `_parse_exec_targets` is a pure function that should have unit tests. The `exec_cmd` command could use CliRunner tests for argument parsing (help output, missing args) and mocked execution tests.

### Self-Review
**Limitations:** Could not verify whether `AutomationContext(inventory=inv_source)` correctly accepts an `Inventory` object vs only dicts — the `__init__` observation hit `AutomationError.__init__` instead of `AutomationContext.__init__`. The agent search confirmed the parameters but I couldn't directly see the `inventory` parameter's type annotation or handling. Also could not see the full Ansible `shell` module parameter list to confirm whether `timeout` is an accepted parameter.

### Feature Requests
- Observation for `__init__` should correctly resolve when the target class contains inner classes or exceptions with their own `__init__` — the observation system returned `AutomationError.__init__` instead of `AutomationContext.__init__`.
- Include callers of new functions/methods to verify nothing else references them unexpectedly.
- For CLI commands, automatically check if the command appears in help text or is registered with the CLI group.

## gemini [BLOCK]

### src/ftl2/cli.py (Imports)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Importing `HostConfig` from `ftl2.types` is correct and necessary for typing the new helper function and command.

### src/ftl2/cli.py:_parse_exec_targets
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

There are multiple edge cases in target parsing:
1. **IPv6 Parsing Bug**: For targets specified as IPv6 addresses without brackets (e.g. `2001:db8::1`), `":" in target` is `True`, causing `rsplit(":", 1)` to incorrectly split it into host `2001:db8::` and port `1`. This breaks IPv6 support entirely unless brackets are strictly used.
2. **Missing Post-Split Validation**: If a malformed target is passed like `user@`, `rsplit` splits it into `user` and `""`. No check is performed afterwards, leading to appending a `HostConfig` with an empty string name and host.
3. **IPv6 Brackets**: Standard SSH connections often expect bracketless IP addresses as the hostname. Target strings with bracketed IPv6 addresses (e.g., `[2001:db8::1]:22`) will leave the square brackets in `ansible_host` which can cause SSH connection failures.

### src/ftl2/cli.py:exec_cmd
**Verdict:** BLOCK
**Correctness:** BROKEN
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** PARTIAL

1. **Silent Connection/Execution Failures on Multi-Host Runs (CRITICAL)**: Under `multi = True` (multiple targets), if a host execution fails with an exception (such as DNS failure, connection timeout, or host unreachable), `stdout` and `stderr` will both be empty strings. The code only prints output if `stdout` or `stderr` splitlines have content. Because the error-reporting fallback `click.echo(result.error or "Command failed")` is only placed in the `else` branch (single host execution), multi-host runs will silently fail to output any error messages for failed hosts. The command will exit with an error code, but the console will be completely silent about what failed.
2. **Missing Inventory Group Support**: If an inventory `-i` file is provided, `exec` only verifies `name in all_hosts` (hosts dictionary). It does not check group names (e.g. `webservers` or `@webservers`), even though targeting group names is a core pattern in FTL2 and Ansible inventories. This causes a `ClickException` when users try to target a group.
3. **Lack of Automated Tests**: A brand new user-facing CLI subcommand (`exec`) has been introduced with absolutely zero test coverage in `tests/test_cli.py` or elsewhere.

### Self-Review
**Limitations:** - Could not run execution end-to-end to verify SSH connection behavior with bracketed IPv6 addresses directly since no remote SSH environment is active.

### Feature Requests
- Include capability to automatically run the corresponding test suite and report test results in the prompt payload to verify if changes broke any existing tests.
- Show code context surrounding any referenced classes/types (e.g., `HostConfig`, `ExecuteResult`) to make it easier to verify method signatures and properties without additional searches.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#161
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/cli.py:_parse_exec_targets
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The parsing logic has two edge cases worth noting:

1. **No validation on port**: `int(port_str)` will raise `ValueError` for non-numeric ports, which is caught by the caller — this is fine. But negative ports or ports > 65535 are silently accepted.

2. **IPv6 addresses**: Targets like `user@[::1]:22` will break because `rsplit(":", 1)` will split on the IPv6 colon, not the port separator. This is a known pattern limitation but worth documenting or handling since sysadmins do use IPv6.

3. **No unit tests** for `_parse_exec_targets`. This is a pure function with clear inputs/outputs — ideal for unit testing. Cases to cover: basic hostname, user@host, host:port, user@host:port, comma-separated, empty elements, invalid port.

### src/ftl2/cli.py:exec_cmd
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

1. **Dead import**: `import sys` at line 2224 is unused — leftover from development.

2. **`timeout` is passed as a module param, not a connection timeout**: The `--timeout` option (default 300s) is passed as `timeout=timeout` to `ftl.run_on()` kwargs, which become module params (`**params`). The `shell` module does accept a `timeout` param for command timeout, so this works — but the semantics may surprise users who expect it to control SSH connection timeout. The help text says "Command timeout" which is accurate.

3. **Exit code handling for multi-host**: When multiple hosts fail, only the first failure's `rc` is used as the exit code (line 2313-2315: `if exit_code == 0: exit_code = rc`). This matches Ansible's behavior and is reasonable, but it means if host A succeeds and host B fails with rc=2, the overall exit is 2. Fine.

4. **`result.output` could be empty dict on error**: The code does `result.output.get("stdout", "")` and `result.output.get("rc", 1)` which handles this correctly via `.get()` with defaults.

5. **`become_overrides` only sent when `--become` flag is set**: This is correct — without `--become`, `become_overrides` is `None` and remote execution proceeds without privilege escalation.

6. **Inventory path handling when `--inventory` is provided with `--user`**: The `--user` flag is ignored when `--inventory` is provided (the inventory's `ansible_user` is used instead). This isn't documented in the help text — could confuse users who pass both.

7. **The `nl=not stdout.endswith("\n")` pattern** (lines 2306, 2308): This prevents double-newlines when the output already ends with `\n`. Smart detail.

8. **No tests at all** for the `exec` command. The test file has no `exec`-related test cases. At minimum, testing `_parse_exec_targets` as a unit test and the CLI help/argument validation would be valuable. The integration flow with `AutomationContext` is harder to test but mockable.

### Self-Review
**Limitations:** Could not see the full `_execute_on_host` method to verify how `_become_overrides` is consumed downstream. Could not verify whether the dict-based inventory structure passed to `AutomationContext` (the `{"all": {"hosts": ...}}` literal) is correctly parsed by the inventory loading code — this path may only be tested via integration tests. The test file observation was truncated at 3037 lines so some exec-related tests could exist beyond the truncation point, though the grep for "exec" found nothing.

### Feature Requests
- Include the `__init__` signature of key classes (like `AutomationContext`) in observations when the diff calls them, so reviewers can verify argument compatibility without extra reads
- When a function is added, automatically grep the test files for test coverage of that specific function and include in observations

## gemini [CONCERN]

### src/ftl2/cli.py:_parse_exec_targets
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

- **IPv6 Parsing Limitation:** The helper splits ports using `rsplit(":", 1)`. If an unbracketed IPv6 address (e.g., `2001:db8::1`) is supplied without a port, it will incorrectly split at the last colon and attempt to parse the last hex/digit block (`1`) as the port. Bracketed IPv6 targets (e.g. `[2001:db8::1]:2222`) work correctly.
- **No Unit Tests:** There are no tests added in `tests/test_cli.py` or elsewhere to verify that target parsing behaves as expected under different inputs (e.g. usernames, ports, commas, invalid syntax).

### src/ftl2/cli.py:exec_cmd
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

- **SSH Fallback for Localhost:** When executing a command on `localhost` without an explicit `--inventory` file, `ansible_connection` defaults to `"ssh"`. This forces an SSH connection to `127.0.0.1` rather than running locally (unless the user uses `--inventory` and specifies `ansible_connection: local`).
- **No Command Tests:** No CLI integration tests are provided. A test using Click's `CliRunner` should be added to `tests/test_cli.py` to verify command registration, argument handling, output prefixes for multi-host scenarios, and proper exit code propagation upon failure.
- **Clean Execution Output:** The output formatting handles single-host and multi-host scenarios cleanly, formatting and prefixing lines beautifully. Exception and failure exit codes are propagated accurately.

### Self-Review
**Limitations:** None. All production implementations and related CLI test files were fully accessible to verify class configurations, properties, and current test coverage.

PR #167, opened 2026-07-06
Add native GCP Compute Engine modules

benthomasson:
## Summary
- Add `ftl_gcp_compute_instance()` and `ftl_gcp_compute_instance_info()` native FTL modules for GCP Compute Engine
- Uses `google-cloud-compute` SDK with `asyncio.to_thread()` wrapping (SDK is sync-only)
- Full state management: create/start/stop/delete with idempotent name-based lookup and operation polling
- Registered in all 4 places: `gcp/__init__.py`, `ftl_modules/__init__.py`, `executor.py`, `pyproject.toml` (`gcp` extra)
- Follows the same pattern as the existing `aws/ec2.py` modules

## Test plan
- [x] `uv pip install -e ".[gcp]"` installs successfully
- [x] Module registration verified via `get_module('gcp_compute_instance')` and FQCN
- [x] Executor resolves both short names and FQCNs
- [x] 1156 existing tests pass (1 pre-existing failure unrelated)
- [x] End-to-end: created, verified, and deleted a GCE instance via ftl2-agentcraft

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ gcp = [
     "google-cloud-run>=0.10.0",
     "google-cloud-artifact-registry>=1.0.0",
     "google-cloud-secret-manager>=2.0.0",
+    "google-cloud-compute>=1.14.0",
     "google-auth>=2.0.0",
 ]
 vault = [
--- a/src/ftl2/ftl_modules/__init__.py
+++ b/src/ftl2/ftl_modules/__init__.py
@@ -29,6 +29,8 @@ from ftl2.ftl_modules.aws import (
 from ftl2.ftl_modules.gcp import (
     ftl_artifact_registry_repository,
     ftl_cloud_run_service,
+    ftl_gcp_compute_instance,
+    ftl_gcp_compute_instance_info,
     ftl_secret_manager_secret,
 )
 from ftl2.ftl_modules.command import ftl_command, ftl_shell
@@ -76,6 +78,8 @@ FTL_MODULES: dict[str, ModuleFunc] = {
     "cloud_run_service": ftl_cloud_run_service,
     "artifact_registry_repository": ftl_artifact_registry_repository,
     "secret_manager_secret": ftl_secret_manager_secret,
+    "gcp_compute_instance": ftl_gcp_compute_instance,
+    "gcp_compute_instance_info": ftl_gcp_compute_instance_info,
 }
 
 # Maps Ansible FQCNs to FTL module functions for compatibility
@@ -101,6 +105,8 @@ ANSIBLE_COMPAT: dict[str, ModuleFunc] = {
     "google.cloud.cloud_run_service": ftl_cloud_run_service,
     "google.cloud.artifact_registry_repository": ftl_artifact_registry_repository,
     "google.cloud.secret_manager_secret": ftl_secret_manager_secret,
+    "google.cloud.gcp_compute_instance": ftl_gcp_compute_instance,
+    "google.cloud.gcp_compute_instance_info": ftl_gcp_compute_instance_info,
 }
 
 
@@ -197,4 +203,6 @@ __all__ = [
     "ftl_cloud_run_service",
     "ftl_artifact_registry_repository",
     "ftl_secret_manager_secret",
+    "ftl_gcp_compute_instance",
+    "ftl_gcp_compute_instance_info",
 ]
--- a/src/ftl2/ftl_modules/executor.py
+++ b/src/ftl2/ftl_modules/executor.py
@@ -142,6 +142,7 @@ def _get_module(name: str) -> Any:
     # Import here to avoid circular import
     from ftl2.ftl_modules.aws.ec2 import ftl_ec2_instance, ftl_ec2_instance_info
     from ftl2.ftl_modules.aws.route53 import ftl_route53_info, ftl_route53_record
+    from ftl2.ftl_modules.gcp.compute import ftl_gcp_compute_instance, ftl_gcp_compute_instance_info
     from ftl2.ftl_modules.command import ftl_command, ftl_shell
     from ftl2.ftl_modules.gcp.artifact_registry import ftl_artifact_registry_repository
     from ftl2.ftl_modules.gcp.cloud_run import ftl_cloud_run_service
@@ -167,6 +168,8 @@ def _get_module(name: str) -> Any:
         "ec2_instance_info": ftl_ec2_instance_info,
         "route53_record": ftl_route53_record,
         "route53_info": ftl_route53_info,
+        "gcp_compute_instance": ftl_gcp_compute_instance,
+        "gcp_compute_instance_info": ftl_gcp_compute_instance_info,
         "dnf": ftl_dnf,
         "dnf5": ftl_dnf,
         # FQCN mappings
@@ -188,9 +191,13 @@ def _get_module(name: str) -> Any:
         "cloud_run_service": ftl_cloud_run_service,
         "artifact_registry_repository": ftl_artifact_registry_repository,
         "secret_manager_secret": ftl_secret_manager_secret,
+        "gcp_compute_instance": ftl_gcp_compute_instance,
+        "gcp_compute_instance_info": ftl_gcp_compute_instance_info,
         "google.cloud.cloud_run_service": ftl_cloud_run_service,
         "google.cloud.artifact_registry_repository": ftl_artifact_registry_repository,
         "google.cloud.secret_manager_secret": ftl_secret_manager_secret,
+        "google.cloud.gcp_compute_instance": ftl_gcp_compute_instance,
+        "google.cloud.gcp_compute_instance_info": ftl_gcp_compute_instance_info,
     }
     return modules.get(name)
 
--- a/src/ftl2/ftl_modules/gcp/__init__.py
+++ b/src/ftl2/ftl_modules/gcp/__init__.py
@@ -5,10 +5,16 @@ Async Google Cloud modules using the Google Cloud Python SDK.
 
 from ftl2.ftl_modules.gcp.artifact_registry import ftl_artifact_registry_repository
 from ftl2.ftl_modules.gcp.cloud_run import ftl_cloud_run_service
+from ftl2.ftl_modules.gcp.compute import (
+    ftl_gcp_compute_instance,
+    ftl_gcp_compute_instance_info,
+)
 from ftl2.ftl_modules.gcp.secret_manager import ftl_secret_manager_secret
 
 __all__ = [
     "ftl_cloud_run_service",
     "ftl_artifact_registry_repository",
     "ftl_secret_manager_secret",
+    "ftl_gcp_compute_instance",
+    "ftl_gcp_compute_instance_info",
 ]
--- a/src/ftl2/ftl_modules/gcp/compute.py
+++ b/src/ftl2/ftl_modules/gcp/compute.py
@@ -1,0 +1,437 @@
+"""FTL GCP Compute Engine module.
+
+Async GCE instance management using google-cloud-compute.
+All SDK calls are synchronous and wrapped with asyncio.to_thread().
+"""
+
+import asyncio
+from typing import Any
+
+from ftl2.ftl_modules.exceptions import FTLModuleError, requires_extra
+
+__all__ = ["ftl_gcp_compute_instance", "ftl_gcp_compute_instance_info"]
+
+_RUNNING_STATES = ("RUNNING",)
+_STOPPED_STATES = ("TERMINATED", "STOPPED")
+_ACTIVE_STATES = ("RUNNING", "STAGING", "PROVISIONING", "STOPPING", "SUSPENDING", "SUSPENDED",
+                   "REPAIRING", "TERMINATED", "STOPPED")
+
+
+def _extract_instance(instance: Any) -> dict[str, Any]:
+    """Normalize a GCP Instance protobuf to a flat result dict."""
+    external_ip = None
+    internal_ip = None
+    network = None
+    subnet = None
+
+    if instance.network_interfaces:
+        ni = instance.network_interfaces[0]
+        internal_ip = ni.network_i_p if ni.network_i_p else None
+        network = ni.network if ni.network else None
+        subnet = ni.subnetwork if ni.subnetwork else None
+        if ni.access_configs:
+            for ac in ni.access_configs:
+                if ac.nat_i_p:
+                    external_ip = ac.nat_i_p
+                    break
+
+    tags = list(instance.tags.items) if instance.tags and instance.tags.items else []
+
+    metadata = {}
+    if instance.metadata and instance.metadata.items:
+        for item in instance.metadata.items:
+            metadata[item.key] = item.value
+
+    zone_name = instance.zone.rsplit("/", 1)[-1] if instance.zone else ""
+    machine_type_short = instance.machine_type.rsplit("/", 1)[-1] if instance.machine_type else ""
+
+    return {
+        "instance_id": str(instance.id) if instance.id else "",
+        "name": instance.name,
+        "machine_type": machine_type_short,
+        "status": instance.status,
+        "external_ip": external_ip,
+        "internal_ip": internal_ip,
+        "zone": zone_name,
+        "network": network,
+        "subnet": subnet,
+        "self_link": instance.self_link,
+        "tags": tags,
+        "metadata": metadata,
+        "creation_timestamp": instance.creation_timestamp or "",
+    }
+
+
+async def _find_instance(client: Any, project: str, zone: str, name: str) -> Any | None:
+    """Find an existing instance by name, return None if not found."""
+    from google.api_core.exceptions import NotFound
+
+    try:
+        return await asyncio.to_thread(client.get, project=project, zone=zone, instance=name)
+    except NotFound:
+        return None
+
+
+async def _wait_for_operation(
+    project: str, zone: str, operation: str, timeout: int,
+) -> None:
+    """Wait for a zone operation to complete."""
+    from google.cloud import compute_v1
+
+    ops_client = compute_v1.ZoneOperationsClient()
+    waited = 0
+    interval = 5
+    while waited < timeout:
+        result = await asyncio.to_thread(
+            ops_client.get, project=project, zone=zone, operation=operation,
+        )
+        if result.status == compute_v1.Operation.Status.DONE:
+            if result.error and result.error.errors:
+                msgs = [e.message for e in result.error.errors]
+                raise FTLModuleError(
+                    f"Operation {operation} failed: {'; '.join(msgs)}",
+                    operation=operation,
+                )
+            return
+        await asyncio.sleep(interval)
+        waited += interval
+
+    raise FTLModuleError(
+        f"Timed out waiting for operation {operation} after {timeout}s",
+        operation=operation,
+    )
+
+
+@requires_extra("gcp", "google.cloud.compute_v1")
+async def ftl_gcp_compute_instance(
+    name: str,
+    project: str,
+    zone: str = "us-central1-a",
+    state: str = "present",
+    machine_type: str = "e2-micro",
+    source_image: str | None = None,
+    disk_size_gb: int = 20,
+    network: str = "global/networks/default",
+    subnet: str | None = None,
+    tags: list[str] | None = None,
+    metadata: dict[str, str] | None = None,
+    service_account_email: str | None = None,
+    service_account_scopes: list[str] | None = None,
+    labels: dict[str, str] | None = None,
+    wait: bool = True,
+    wait_timeout: int = 600,
+    **kwargs: Any,
+) -> dict[str, Any]:
+    """Manage GCP Compute Engine instances.
+
+    Args:
+        name: Instance name (used for idempotent lookup)
+        project: GCP project ID
+        zone: GCE zone (e.g., us-central1-a)
+        state: Desired state (present, running, started, stopped, terminated, absent)
+        machine_type: Machine type (e.g., e2-micro, n1-standard-1)
+        source_image: Boot disk source image (e.g., projects/fedora-cloud/global/images/family/fedora-cloud-42)
+        disk_size_gb: Boot disk size in GB (default 20)
+        network: VPC network (default: global/networks/default)
+        subnet: Subnet (optional)
+        tags: Network tags
+        metadata: Instance metadata key-value pairs
+        service_account_email: Service account email
+        service_account_scopes: Service account scopes
+        labels: Labels to apply
+        wait: Wait for state transitions (default True)
+        wait_timeout: Timeout in seconds (default 600)
+        **kwargs: Additional parameters
+
+    Returns:
+        Result dict with changed, instance_id, state, and instance details
+    """
+    from google.api_core.exceptions import GoogleAPICallError
+    from google.cloud import compute_v1
+
+    state = state.lower()
+    if state == "started":
+        state = "running"
+    if state == "absent":
+        state = "terminated"
+    if state == "present":
+        state = "running"
+
+    client = compute_v1.InstancesClient()
+
+    try:
+        existing = await _find_instance(client, project, zone, name)
+
+        if state == "running":
+            if existing:
+                current = existing.status
+                if current in _RUNNING_STATES:
+                    return {
+                        "changed": False,
+                        "instance_id": str(existing.id),
+                        "state": current,
+                        "instance": _extract_instance(existing),
+                    }
+                if current in ("STAGING", "PROVISIONING"):
+                    if wait:
+                        waited = 0
+                        while waited < wait_timeout:
+                            inst = await asyncio.to_thread(
+                                client.get, project=project, zone=zone, instance=name,
+                            )
+                            if inst.status in _RUNNING_STATES:
+                                return {
+                                    "changed": False,
+                                    "instance_id": str(inst.id),
+                                    "state": inst.status,
+                                    "instance": _extract_instance(inst),
+                                }
+                            await asyncio.sleep(5)
+                            waited += 5
+                        raise FTLModuleError(
+                            f"Timed out waiting for {name} to reach RUNNING after {wait_timeout}s",
+                            instance=name,
+                        )
+                    return {
+                        "changed": False,
+                        "instance_id": str(existing.id),
+                        "state": current,
+                        "instance": _extract_instance(existing),
+                    }
+                if current in ("STOPPING", "SUSPENDING"):
+                    waited = 0
+                    while waited < wait_timeout:
+                        inst = await asyncio.to_thread(
+                            client.get, project=project, zone=zone, instance=name,
+                        )
+                        if inst.status in _STOPPED_STATES:
+                            break
+                        await asyncio.sleep(5)
+                        waited += 5
+                    current = "TERMINATED"
+                if current in _STOPPED_STATES:
+                    op = await asyncio.to_thread(
+                        client.start, project=project, zone=zone, instance=name,
+                    )
+                    if wait:
+                        await _wait_for_operation(project, zone, op.name, wait_timeout)
+                    inst = await asyncio.to_thread(
+                        client.get, project=project, zone=zone, instance=name,
+                    )
+                    return {
+                        "changed": True,
+                        "instance_id": str(inst.id),
+                        "state": inst.status,
+                        "instance": _extract_instance(inst),
+                    }
+                raise FTLModuleError(
+                    f"Instance {name} is in unexpected state: {current}",
+                    instance=name,
+                    current_state=current,
+                )
+
+            if not source_image:
+                raise FTLModuleError(
+                    "source_image is required to create a new instance",
+                )
+
+            machine_type_url = f"zones/{zone}/machineTypes/{machine_type}"
+
+            boot_disk = compute_v1.AttachedDisk(
+                auto_delete=True,
+                boot=True,
+                initialize_params=compute_v1.AttachedDiskInitializeParams(
+                    source_image=source_image,
+                    disk_size_gb=disk_size_gb,
+                ),
+            )
+
+            access_config = compute_v1.AccessConfig(
+                name="External NAT",
+                type_="ONE_TO_ONE_NAT",
+            )
+            network_interface = compute_v1.NetworkInterface(
+                network=network,
+                access_configs=[access_config],
+            )
+            if subnet:
+                network_interface.subnetwork = subnet
+
+            instance_resource = compute_v1.Instance(
+                name=name,
+                machine_type=machine_type_url,
+                disks=[boot_disk],
+                network_interfaces=[network_interface],
+            )
+
+            if tags:
+                instance_resource.tags = compute_v1.Tags(items=tags)
+
+            if metadata:
+                instance_resource.metadata = compute_v1.Metadata(
+                    items=[
+                        compute_v1.Items(key=k, value=v)
+                        for k, v in metadata.items()
+                    ],
+                )
+
+            if labels:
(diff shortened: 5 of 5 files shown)


benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#167
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/ftl_modules/gcp/compute.py
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Three correctness issues:

1. **Line 211 — hardcoded `current = "TERMINATED"` after STOPPING/SUSPENDING wait loop.** After the while loop (lines 201-210), `current` is unconditionally set to `"TERMINATED"` regardless of whether the loop actually broke (instance reached a stopped state) or timed out. If the loop times out, the code falls through to `if current in _STOPPED_STATES` (line 212) — which is True because `"TERMINATED"` is in `_STOPPED_STATES` — and issues a `client.start` call on an instance that may still be in STOPPING state. The correct fix: after the loop, re-read the actual status, or raise on timeout, as the EC2 module does with its `_wait_for_state` pattern. Also, the variable should be set to the actual observed state (e.g., `inst.status`), not a hardcoded string.

2. **Line 211 — SUSPENDING state falls through to start.** A SUSPENDING instance transitions to SUSPENDED, not TERMINATED/STOPPED. The code waits for `_STOPPED_STATES` (`"TERMINATED"`, `"STOPPED"`), but SUSPENDED is not in that set — so the wait loop will always time out for SUSPENDING instances, then hardcode `current = "TERMINATED"` and attempt to start an instance that is actually SUSPENDED. The `client.start` call may fail or behave unexpectedly on a SUSPENDED instance (GCP requires `resume`, not `start`, for suspended instances).

3. **Lines 420-422 — pagination not fully async.** `client.list()` returns a pager object. Iterating the pager with `for inst in page_result` triggers synchronous HTTP calls for subsequent pages on the main thread. Only the initial `client.list` call is wrapped in `to_thread`. For large instance lists this blocks the event loop. The EC2 module avoids this with a proper paginator pattern. Consider wrapping the iteration in `to_thread` as well, or use `await asyncio.to_thread(lambda: list(client.list(request=request)))`.

4. **`_ACTIVE_STATES` is defined but never used** (line 16-17) — minor dead code.

### src/ftl2/ftl_modules/gcp/compute.py:_wait_for_operation
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Creates a fresh `ZoneOperationsClient` per call, polls correctly, checks for operation errors, raises on timeout. Follows a reasonable pattern. Minor nit: the client is created inside the function every call rather than being passed in, but this matches the sync-wrapped pattern used throughout.

### src/ftl2/ftl_modules/gcp/compute.py:_find_instance
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Catches only `NotFound`, consistent with the narrowed catch pattern applied to the AWS `_find_instance` in commit 3c4992d. Clean and correct.

### src/ftl2/ftl_modules/gcp/compute.py:_extract_instance
**Verdict:** PASS
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Generally correct. The attribute names `network_i_p` and `nat_i_p` look unusual but are correct for the google-cloud-compute protobuf-generated API (the proto field `network_IP` becomes `network_i_p` in the Python client). The function handles missing fields defensively. One minor note: `instance.id` can be `0` for very early instances (unlikely in practice) — `if instance.id` would treat `0` as falsy and return `""`, though GCE IDs are large integers so this is theoretical.

### src/ftl2/ftl_modules/gcp/compute.py:ftl_gcp_compute_instance_info
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The pagination issue described above (lines 420-422) can block the event loop. The `NotFound` exception is caught inside the inner try but it's also a subclass of `GoogleAPICallError`, so the outer except would also catch it — however the inner try/except handles it first, so there's no functional bug. The `filter` attribute assignment on line 418 shadows a Python builtin but this is on a request object so it's fine.

### src/ftl2/ftl_modules/gcp/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Mirrors the `aws/__init__.py` structure exactly. Clean re-exports.

### src/ftl2/ftl_modules/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** PARTIAL
**Integration:** WIRED

GCP modules properly added to `FTL_MODULES`, `ANSIBLE_COMPAT`, and `__all__`. Registration follows the same pattern as AWS modules. However, the existing tests in `test_ftl_modules.py` do not check for the new `gcp_compute_instance` entries in `FTL_MODULES` or `google.cloud.gcp_compute_instance` in `ANSIBLE_COMPAT`.

### src/ftl2/ftl_modules/executor.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

GCP modules added to both short-name and FQCN mappings inside `_get_module()`, consistent with the existing AWS registration pattern. Import is deferred inside the function to avoid circular imports, matching the existing pattern.

### Self-Review
**Limitations:** Could not install/import `google-cloud-compute` to verify protobuf attribute names (`network_i_p`, `nat_i_p`) or confirm `client.start` vs `client.resume` behavior for SUSPENDED instances. Review of pagination behavior is based on documented behavior of the google-cloud-compute library, not empirical verification. No test files exist for either AWS or GCP modules, so I could not check for broken test assertions.

### Feature Requests
- Include the full existing test files that reference modified modules (e.g. `test_ftl_modules.py`) in the diff or observations, so the reviewer can verify whether existing assertions break
- When a new module follows an existing pattern (like GCP mirroring AWS), show the reference module's full source as an observation to facilitate pattern-comparison review

## gemini [CONCERN]

### pyproject.toml
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly registers the new `gcp` optional dependency group, specifying `google-cloud-compute>=1.14.0` in a manner consistent with existing optional dependency configurations (e.g., `aws`, `vault`).

### src/ftl2/ftl_modules/__init__.py
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Successfully exposes the new GCP modules in `FTL_MODULES` and standardizes the Ansible FQCN compat mappings for `google.cloud.gcp_compute_instance` and `google.cloud.gcp_compute_instance_info`. Crucially, it avoids top-level imports of optional dependencies, ensuring that importing `ftl2.ftl_modules` does not raise an `ImportError` if `google-cloud-compute` is uninstalled.
---

### src/ftl2/ftl_modules/executor.py
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Correctly integrates the GCP module functions into the lazy internal `_get_module` function to prevent circular dependency imports and avoid executing optional imports on startup.
---

### src/ftl2/ftl_modules/gcp/__init__.py
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: Standard, clean package initializer exposing the public endpoints.
---

### src/ftl2/ftl_modules/gcp/compute.py:ftl_gcp_compute_instance
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

1. **UNTESTED**: No test files (e.g., `tests/test_gcp_compute.py`) were added to the codebase. An implementation of this size and complexity must have mocked unit tests verifying state transitions and failure modes.
2. **State Machine Timeout Logic Bug**: In the `running` state check block, if the instance is currently `"STOPPING"` or `"SUSPENDING"`, a loop waits for it to finish stopping. However, if this loop times out (`waited >= wait_timeout`), no error is raised; instead, `current` is unconditionally set to `"TERMINATED"`. The code then proceeds to attempt starting the stopping/suspending instance, which will cause an unexpected API exception from Google instead of raising a clean, descriptive module timeout error.

### src/ftl2/ftl_modules/gcp/compute.py:ftl_gcp_compute_instance_info
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

1. **UNTESTED**: Lacks test coverage.
2. **Event Loop Blocking Pager Bug**: In the list operation:
   ```python
   page_result = await asyncio.to_thread(client.list, request=request)
   for inst in page_result:
       instances.append(_extract_instance(inst))
   ```
   The `client.list()` function returns a lazy generator/pager that makes subsequent synchronous HTTP calls to fetch extra pages as they are requested. Iterating over `page_result` in the main async thread with `for inst in page_result:` means any subsequent page fetch will block the async event loop. To resolve this, evaluate the lazy generator inside `asyncio.to_thread`, e.g., `instances_list = await asyncio.to_thread(list, client.list(request=request))`.

### Self-Review
**Limitations:** - Lacked local GCP test credentials to verify actual API communication, but verified logical correctness against the Google Cloud API specifications.
- Checked that standard environmental authentication (`GOOGLE_APPLICATION_CREDENTIALS`) is expected, matching the pattern used in the AWS modules.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#167
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### src/ftl2/ftl_modules/gcp/compute.py:_extract_instance
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Correctly handles GCP protobuf attribute names (`network_i_p`, `nat_i_p` — these are the real Python attribute names from the proto-to-Python name mangling of `networkIP`/`natIP`). Safely handles None/empty for all fields. Extracts zone and machine_type short names from full resource URLs. Follows the same shape as the AWS `_extract_instance` but adapted for GCP's protobuf response objects.

### src/ftl2/ftl_modules/gcp/compute.py:_find_instance
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Catches only `NotFound`, consistent with the narrowed catch pattern in commit 3c4992d for AWS. Other `GoogleAPICallError` subclasses propagate to the caller's outer try/except, which is correct — only a missing instance should return None.

### src/ftl2/ftl_modules/gcp/compute.py:_wait_for_operation
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Creates a new `ZoneOperationsClient()` per call (lightweight, acceptable). Checks `result.error.errors` after DONE status — correct, GCP operations can complete with errors. Timeout tracking matches the AWS `_wait_for_state` pattern.

### src/ftl2/ftl_modules/gcp/compute.py:ftl_gcp_compute_instance (state="running", existing SUSPENDED)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Lines 199-215: When the existing instance is `SUSPENDED`, the code waits for it to transition to a `_STOPPED_STATES` member. But a GCP suspended instance stays `SUSPENDED` indefinitely — it requires an explicit `resume` API call. This loop will always time out after `wait_timeout` seconds. The fix would be to call `client.resume()` for `SUSPENDED` instances, or exclude `SUSPENDED` from this branch and treat it as an error/unsupported transition. This is a real but edge-case bug — most users won't hit it unless they suspend instances.

### src/ftl2/ftl_modules/gcp/compute.py:ftl_gcp_compute_instance (overall)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

State normalization (`started`→`running`, `absent`→`terminated`, `present`→`running`) is correct. The state machine covers create, start-from-stopped, idempotent-already-running, stop, and delete. GCP uses `TERMINATED` as its normal stopped state (unlike AWS), and including both `TERMINATED` and `STOPPED` in `_STOPPED_STATES` is correct. The `@requires_extra("gcp", "google.cloud.compute_v1")` decorator correctly gates on the optional dependency. All google imports are inside function bodies, so the module is importable without the GCP SDK installed. The broad `except Exception` at line 372 is consistent with the AWS modules.

### src/ftl2/ftl_modules/gcp/compute.py:ftl_gcp_compute_instance_info
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Single-instance lookup catches `NotFound` and returns empty list — correct idempotent behavior. List path wraps the GCP pager in `list()` within `to_thread` — correctly drains all pages synchronously inside the thread. The `filter_expr` parameter maps to the GCP API filter syntax. Using `lambda: list(...)` is necessary here since `to_thread` needs a no-arg callable for the synchronous pagination iteration.

### src/ftl2/ftl_modules/gcp/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Mirrors the AWS `__init__.py` structure exactly. The imports from `compute.py` are safe at module load time because `@requires_extra` decorates the functions without importing google packages.

### src/ftl2/ftl_modules/__init__.py (GCP additions)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

GCP modules are registered in all three places: `FTL_MODULES` (short names), `ANSIBLE_COMPAT` (FQCNs), and `__all__`. The FQCN `google.cloud.gcp_compute_instance` matches the Ansible google.cloud collection naming convention.

### src/ftl2/ftl_modules/executor.py (GCP additions)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The `_get_module` function adds GCP modules to both the short-name and FQCN dictionaries, consistent with how AWS modules are registered. Import is inside the function body to avoid circular imports, matching the existing pattern.

### Self-Review
**Limitations:** Could not verify the exact GCP protobuf attribute names (`network_i_p`, `nat_i_p`, etc.) against the actual `google-cloud-compute` library since it's not installed. Could not verify whether the `SUSPENDED` state is actually reachable for standard instances or only for instances with suspend/resume enabled. No AWS test files exist to compare test coverage patterns against.

### Feature Requests
- Include existing test files for sibling modules (e.g., AWS tests if they existed) to assess whether the project has a test coverage norm for cloud modules
- Show the Ansible collection FQCN reference to verify FQCN mappings are correct

## gemini [CONCERN]

### pyproject.toml
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Properly adds `"google-cloud-compute>=1.14.0"` to the `gcp` optional dependency section, matching the existing structure of other optional integrations like `aws` and `vault`.

### src/ftl2/ftl_modules/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Correctly registers `ftl_gcp_compute_instance` and `ftl_gcp_compute_instance_info` in both short-name (`FTL_MODULES`) and FQCN (`ANSIBLE_COMPAT`) mappings. Imports are properly structured so they do not trigger a load of the underlying Google cloud package at startup.

### src/ftl2/ftl_modules/executor.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Integrates the GCP modules using lazy-imports inside `_get_module` function scope. This prevents circular dependencies and guarantees that the Google SDK is only loaded when a GCP-specific task is executed.

### src/ftl2/ftl_modules/gcp/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Provides a clean package entrypoint for GCP modules, properly declaring exported functions in `__all__`.

### src/ftl2/ftl_modules/gcp/compute.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

1. **Design and Non-Blocking Operations**: Using `asyncio.to_thread` for all synchronous SDK calls is an excellent implementation choice, ensuring non-blocking execution in the main async event loop.
2. **Control Flow**: The state machine logic in `ftl_gcp_compute_instance` (handling state transitions from staging, provisioning, stopping, stopped, and running) is highly elegant, particularly the use of Python's `while-else` construct.
3. **Pagination Offloading**: Forcing list construction (`list(client.list(...))`) inside the lambda within `asyncio.to_thread` is outstanding because it prevents lazy-fetching network calls from blocking the main event loop during iteration.
4. **Critical Gap**: There are zero tests provided for this module. Mock-based unit/integration tests must be added to verify behavior (e.g., mock client.get/list/insert/stop/delete and zone operations client) under various VM states before this can be safely merged.

### Self-Review
**Limitations:** - Local unit test runner execution was disabled by system policy, preventing empirical verification of the whole test suite's current pass/fail status. However, static code analysis confirmed that no GCP test files are present.
- Real GCP credentials are not available, so actual runtime validation must rely on mock/fixture-based tests.

### Feature Requests
- Implement standard mock test cases for `ftl_gcp_compute_instance` and `ftl_gcp_compute_instance_info` in `tests/test_gcp_compute.py` using `unittest.mock` or `pytest-mock` to test all major state transitions (present/running, stopped, terminated, etc.) without requiring live cloud environments or API keys.

benthomasson:
# Code Review Report

**Branch:** benthomasson/ftl2#167
**Models:** claude, gemini
**Gate:** [CONCERN] CONCERN

## claude [CONCERN]

### pyproject.toml
VERDICT: PASS
CORRECTNESS: VALID
SPEC_COMPLIANCE: N/A
ISSUE_COMPLIANCE: N/A
BELIEF_COMPLIANCE: N/A
TEST_COVERAGE: N/A
INTEGRATION: WIRED
REASONING: The `gcp` optional dependency group follows the same pattern as the existing `aws` extra. `google-cloud-compute>=1.14.0` is the correct package for the GCE SDK. Properly placed between `aws` and `vault` alphabetically.
---

### src/ftl2/ftl_modules/gcp/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Mirrors the `aws/__init__.py` structure exactly. Exports the two public functions and declares `__all__`.

### src/ftl2/ftl_modules/gcp/compute.py:_extract_instance
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Correctly navigates the GCP protobuf structure. The property names `network_i_p` and `nat_i_p` are the correct python-ified names from `google-cloud-compute` (the protobuf fields `networkIP` and `natIP` get snake-cased with the `_i_p` suffix). Handles missing network interfaces, tags, and metadata gracefully. Extracts short zone and machine_type names from full resource URLs via `rsplit("/", 1)[-1]`.

### src/ftl2/ftl_modules/gcp/compute.py:_find_instance
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Simple and correct — catches only `NotFound`, which is the right exception for a missing instance in the GCE API. The AWS equivalent catches specific `ClientError` codes; this is the GCP equivalent pattern.

### src/ftl2/ftl_modules/gcp/compute.py:_wait_for_operation
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Creates a new `ZoneOperationsClient()` on every call. This is functionally correct but inefficient — each instantiation pays gRPC channel setup. More importantly, the function only handles **zone** operations, but `client.delete` / `client.insert` can return operations that are **zone** operations, so this works for the current use cases. However, if global operations are ever needed, this would silently fail. The poll loop is correct with `waited < timeout` and 5s intervals. One minor concern: if `timeout` is not divisible by `interval` (5), the actual wait time could be up to 4 seconds less than `wait_timeout`. This matches the AWS module's behavior so it's consistent.

### src/ftl2/ftl_modules/gcp/compute.py:ftl_gcp_compute_instance (state machine)
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The state machine is thorough and handles GCP-specific states (SUSPENDED, SUSPENDING, STAGING, PROVISIONING) that don't exist in AWS. Two concerns:

1. **Fall-through after STOPPING/SUSPENDING wait (lines 214-245):** After waiting for a STOPPING/SUSPENDING instance to settle, if it lands in a STOPPED/TERMINATED state (not SUSPENDED), execution falls through to the `if current in _STOPPED_STATES:` check at line 246. This works because the `current` variable was updated at line 221 to the settled state. However, it's subtle — the `if current == "SUSPENDED"` block at line 231 has its own return, but if current is STOPPED/TERMINATED after the while-else, it falls through to line 246. This is functionally correct but fragile — if someone adds a new state check between lines 245 and 246, it would break.

2. **`state="absent"` maps to `"terminated"` which calls `client.delete`** — this is correct GCE semantics (delete the instance), matching the Ansible convention.

3. **The `stopped` state handler (line 339) doesn't check for transitional states** like STOPPING or SUSPENDING. If the instance is already STOPPING, it will call `client.stop` again, which should be harmless (idempotent) but could be handled more gracefully.

### src/ftl2/ftl_modules/gcp/compute.py:ftl_gcp_compute_instance (instance creation, lines 267-337)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

Instance creation is well-structured. Machine type URL is correctly formatted as `zones/{zone}/machineTypes/{machine_type}`. Boot disk, network interface, tags, metadata, labels, and service account are all properly constructed using `compute_v1` protobuf types. The `**kwargs` parameter is accepted but not forwarded to the GCE API, which is fine — it's consistent with how the AWS module handles extra kwargs (they're swallowed). Note: `compute_v1.Items` is used for metadata items — this is correct for the proto definition.

### src/ftl2/ftl_modules/gcp/compute.py:ftl_gcp_compute_instance_info
**Verdict:** CONCERN
**Correctness:** QUESTIONABLE
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The `list` call at line 454-456 wraps pagination in `list(client.list(request=request))` which materializes all pages into memory. For large zones this could be expensive, but it's the simplest correct approach. The `client.list()` method returns a pager that handles pagination automatically, so `list()` will correctly consume all pages. One concern: the `filter` attribute on `ListInstancesRequest` may conflict with Python's built-in `filter` — but since this is a protobuf message attribute, it should work correctly via the property setter. The observation confirms `is_generator: false`, so the `list()` call is correct.

### src/ftl2/ftl_modules/__init__.py (module registration)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

GCP modules registered in all three places: `FTL_MODULES` (short names), `ANSIBLE_COMPAT` (FQCNs), and `__all__`. The FQCN `google.cloud.gcp_compute_instance` follows the Ansible collection naming convention for GCP modules. Import at top-level is guarded by `@requires_extra` decorator which catches ImportError at call-time, so the import won't fail if `google-cloud-compute` isn't installed — **wait, actually it will**. The top-level import in `__init__.py` line 29-32 imports from `ftl2.ftl_modules.gcp`, which imports from `gcp/compute.py`. The `compute.py` file's top-level only imports `FTLModuleError` and `requires_extra` — the `google.cloud` imports are deferred to inside the function bodies. So this is safe.

### src/ftl2/ftl_modules/executor.py (module dispatch)
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

`_get_module` is the runtime dispatch path. GCP modules added to both short name and FQCN dictionaries, matching the pattern of AWS modules. Imports are inside the function to avoid circular imports, consistent with existing code.

### Self-Review
**Limitations:** - Could not verify `compute_v1.Items` is the correct protobuf class for metadata items (vs `compute_v1.Metadata.Items` or similar) — the GCP SDK was not installed in the environment
- No AWS test file was available to compare test patterns in detail — only saw two test functions via grep
- Could not verify that `google.cloud.gcp_compute_instance` is the canonical Ansible FQCN for GCP compute modules (Ansible's actual GCP collection uses `google.cloud.gcp_compute_instance` so this appears correct)

### Feature Requests
- Include test files that cover the same source module (even if they don't exist yet, flag the gap programmatically in the observation data)
- For cloud SDK modules, include the SDK's type stubs or protobuf definitions so reviewers can verify attribute names like `network_i_p`, `nat_i_p`, `compute_v1.Items` without installing the SDK

## gemini [CONCERN]

### pyproject.toml
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly registers the `gcp` optional dependency extra with `google-cloud-compute>=1.14.0`, matching established optional dependency patterns (e.g. `aws`, `vault`).

### src/ftl2/ftl_modules/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Seamlessly imports and registers the GCP module functions under both short-names and FQCNs (`google.cloud.gcp_compute_instance`, `google.cloud.gcp_compute_instance_info`).

### src/ftl2/ftl_modules/executor.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Correctly registers the dynamic loading mappings within `_get_module()`. Doing this inside the function avoids circular imports and ensures the module is fully decoupled at startup.

### src/ftl2/ftl_modules/gcp/__init__.py
**Verdict:** PASS
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** COVERED
**Integration:** WIRED

Standard, clean package entry point exporting the GCP module functions correctly.

### src/ftl2/ftl_modules/gcp/compute.py
**Verdict:** CONCERN
**Correctness:** VALID
**Spec Compliance:** N/A
**Test Coverage:** UNTESTED
**Integration:** WIRED

The implementation is exceptionally robust and clean:
1. **Asynchronous Delegation:** All synchronous GCP SDK client calls are correctly wrapped in `asyncio.to_thread()`, preventing any event-loop blocking.
2. **Dynamic Paging Consumption:** In `ftl_gcp_compute_instance_info`, `lambda: list(client.list(...))` correctly executes within `asyncio.to_thread()`, keeping page fetching non-blocking.
3. **Lazy/Isolated Imports:** All `google.*` client library imports are scoped inside decorated functions or localized utilities. Because of this, importing `ftl2` does not crash if the `gcp` extra is absent.
4. **Decorator Guarding:** The `@requires_extra("gcp", "google.cloud.compute_v1")` decorator properly prevents runtime invocation crashes if the extra is not installed.

**The Concern:**
There are **no unit tests** written for this module. To comply with the engineering mandates, a test suite must be added. Below is a recommended implementation of `tests/test_gcp_compute.py` using standard mocking to run without GCP credentials.

```python
"""Tests for GCP compute modules."""

import sys
from unittest.mock import AsyncMock, MagicMock, patch
import pytest

# Mock out Google client libraries before importing modules to avoid missing dependencies
mock_compute_v1 = MagicMock()
sys.modules["google"] = MagicMock()
sys.modules["google.cloud"] = MagicMock()
sys.modules["google.cloud.compute_v1"] = mock_compute_v1
sys.modules["google.api_core"] = MagicMock()

# Setup exceptions
from google.api_core.exceptions import NotFound
sys.modules["google.api_core.exceptions"] = MagicMock()

from ftl2.ftl_modules.gcp.compute import (
    ftl_gcp_compute_instance,
    ftl_gcp_compute_instance_info,
)


@pytest.mark.asyncio
async def test_ftl_gcp_compute_instance_info_not_found():
    """Test retrieving info for a non-existent instance."""
    client_instance = mock_compute_v1.InstancesClient.return_value
    client_instance.get.side_effect = NotFound("Not Found")

    result = await ftl_gcp_compute_instance_info(
        project="test-project",
        zone="us-central1-a",
        instance="non-existent",
    )

    assert result["changed"] is False
    assert result["instances"] == []


@pytest.mark.asyncio
async def test_ftl_gcp_compute_instance_info_success():
    """Test retrieving info for an existing instance."""
    mock_inst = MagicMock()
    mock_inst.id = 12345
    mock_inst.name = "test-vm"
    mock_inst.status = "RUNNING"
    mock_inst.network_interfaces = []
    mock_inst.tags.items = ["web", "production"]
    mock_inst.metadata.items = []
    mock_inst.zone = "zones/us-central1-a"
    mock_inst.machine_type = "machineTypes/e2-micro"
    mock_inst.self_link = "http://gcp/test-vm"
    mock_inst.creation_timestamp = "2026-01-01T00:00:00"

    client_instance = mock_compute_v1.InstancesClient.return_value
    client_instance.get.return_value = mock_inst

    result = await ftl_gcp_compute_instance_info(
        project="test-project",
        zone="us-central1-a",
        instance="test-vm",
    )

    assert result["changed"] is False
    assert len(result["instances"]) == 1
    assert result["instances"][0]["name"] == "test-vm"
    assert result["instances"][0]["status"] == "RUNNING"


@pytest.mark.asyncio
async def test_ftl_gcp_compute_instance_present_already_running():
    """Test idempotency when state is 'present' and instance is already running."""
    mock_inst = MagicMock()
    mock_inst.id = 12345
    mock_inst.name = "test-vm"
    mock_inst.status = "RUNNING"
    mock_inst.network_interfaces = []
    mock_inst.tags.items = []
    mock_inst.metadata.items = []
    mock_inst.zone = "zones/us-central1-a"
    mock_inst.machine_type = "machineTypes/e2-micro"
    mock_inst.self_link = "http://gcp/test-vm"
    mock_inst.creation_timestamp = ""

    client_instance = mock_compute_v1.InstancesClient.return_value
    client_instance.get.return_value = mock_inst

    result = await ftl_gcp_compute_instance(
        name="test-vm",
        project="test-project",
        zone="us-central1-a",
        state="present",
    )

    assert result["changed"] is False
    assert result["instance_id"] == "12345"
    assert result["state"] == "RUNNING"
```

### Self-Review
**Limitations:** - Local unit testing execution was denied by environment guardrails/policy. Thus, runtime and import verifications were evaluated purely through rigorous manual static analysis.
- The `google-cloud-compute` client-library internal properties (e.g., `network_i_p`, `nat_i_p`) were verified against SDK documentation schemas since local mock validations could not be run.

PR #177, opened 2026-07-27
Default verbose=True and add log_file for AI-first observability

benthomasson:
## Summary
- Change verbose default from False to True in automation() and AutomationContext
- Add log_file parameter (default "ftl2.log") that configures Python logging to file at DEBUG level
- Pass log_file=None to disable automatic logging configuration

AI agents assume success when there is no output. Verbose output and debug logging by default give agents the context they need to diagnose failures.

## Test plan
- [x] Existing tests updated for verbose=True default (158 pass)
- [x] 5 new tests for log_file parameter (default, None, custom path, via automation(), disabled via automation())
- [x] Full test suite passes (163 tests)

### DIFF
--- a/examples/06-automation-context/example_auto_summary.py
+++ b/examples/06-automation-context/example_auto_summary.py
@@ -1,0 +1,47 @@
+#!/usr/bin/env python3
+"""Example: Automatic result collection and summary.
+
+Shows that AutomationContext automatically:
+1. Collects all results (success and failure)
+2. Prints a per-host summary on exit (changed/ok/failed)
+3. Prints error details on exit
+
+No manual error checking needed — just run your tasks and
+the context manager handles the rest.
+
+Run with: uv run python example_auto_summary.py
+"""
+
+import asyncio
+import tempfile
+
+from ftl2 import automation
+
+
+async def main():
+    with tempfile.TemporaryDirectory() as tmpdir:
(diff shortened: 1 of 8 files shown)


benthomasson (commented):
Looks good overall — this addresses real pain points we hit repeatedly in deploy scripts.

One concern: `configure_logging()` is called in `AutomationContext.__init__`, so creating multiple contexts (common in tests, or if someone does sequential deploys) could stack file handlers. The `test_log_file_none_skips_logging` tolerance of `handler_count_after <= handler_count_before + 1` hints at this.

Suggestion: use the `"ftl2"` logger instead of root, tag the handler, and remove-before-add to make it idempotent:

```python
def configure_logging(log_file, file_level=logging.DEBUG):
    logger = logging.getLogger("ftl2")
    # Remove any previous ftl2 file handler
    for h in logger.handlers[:]:
        if getattr(h, "_ftl2_log_file", False):
            logger.removeHandler(h)
            h.close()
    handler = logging.FileHandler(log_file)
    handler.setLevel(file_level)
    handler._ftl2_log_file = True
    handler.setFormatter(logging.Formatter(
        "%(asctime)s %(name)s %(levelname)s %(message)s"
    ))
    logger.addHandler(handler)
    logger.setLevel(file_level)
```

10 contexts → still 1 handler, and it doesn't pollute other libraries' logging.

